Issue #17, opened 2017-02-11
Vagrant - Migrate migrations and load initial data


heymonkeyriot:
The vagrant/provsion.sh file appears to be doing everything correctly

```
# Run syncdb/migrate/load_initial_data/update_index
su - vagrant -c "$PYTHON $PROJECT_DIR/manage.py migrate --noinput && \
                 $PYTHON $PROJECT_DIR/manage.py load_initial_data && \
                 $PYTHON $PROJECT_DIR/manage.py update_index"
```

But none of those commands are being run in the initial `vagrant up` command nor if you run a subsequent `vagrant provision`

tomdyson:
I think it would be better to remove these from `provision.sh` and make the user do them after provisioning. It keeps things consistent with the non-Vagrant instructions.

heymonkeyriot:
Yes there's something to be said for making that consistent. Would you thinking be then that requirements would also be installed manually by the user rather than the provisioning shell file?

One bonus advantage would be that the user would never have difficulties if they hit [the failed to mount folders vagrant problem](http://stackoverflow.com/questions/22717428/vagrant-error-failed-to-mount-folders-in-linux-guest) that can sometimes happen since

daaray:
Interesting.  The provision script *does* create the database....it just creates it in the wrong place.

If you `vagrant ssh` and poke around after a `vagrant up`, the `bakerydemodb` file is located at:

`/home/vagrant`

It needs to be created in `/vagrant`

With regard to moving the commands to manual steps, I would argue that we fix the issue at hand and automate as much as possible via the Vagrant path.  The demo serves two different audiences.  Developers are more than happy to manually execute commands, but this is also a very useful tool for non-developers that can be spun up to showcase Wagtail to potential stakeholders.  For that audience, automation is key and removing the potential for human error will allow for a broader audience to come to the table.

daaray:
Also, just an FYI, the `load_initial_data` currently only loads `bakerydemo.json`.  At the moment, that only provides 1 superuser and no other content.

shacker:
There's lots of work to do on initial data loading - I'll take care of that. If anything special has to be done for vagrant data loading, please wait until I've got it working for the standard instructions.

daaray:
Closing this, as the only remaining work is `load_initial_data` which is handled via #10

PR #20, opened 2017-02-11
#17, add separate vagrant settings file, small fixes elsewhere

daaray:
@tomdyson @heymonkeyriot for your consideration, based off my comments in #17 

 * Adds separate Vagrant settings file
 * corrects issue with Vagrant provision and misplaced db
 * Separate db for vagrant


daaray:
This is not quite ready yet; need to sort out the settings files better.

PR #26, opened 2017-02-12
Clean up Heroku deploy

### LINKED ISSUE #3: Reconcile Heroku deploy bits
The Heroku assets were hastily shoved in; we should review and ensure that it is set up well.

daaray:
Fixes #3 

Breaking out S3 static/media support into a separate ticket as it requires additional configuration.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
+release: pip install -r requirements_heroku.txt
 release: yes "yes" | python manage.py migrate
 web: gunicorn bakerydemo.heroku_wsgi --log-file -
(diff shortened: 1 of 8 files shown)


heymonkeyriot:
This all looks good from my perspective. Should we add something to the readme as well though? Since the Heroku requirements won't get loaded till someone runs `pip install -r requirements/heroku.txt` it'd be good to make that explicit. Perhaps we should add a **Deployment to Heroku** bit to the readme?

daaray:
@heymonkeyriot thanks for the feedback.  Will adjust accordingly.

Issue #24, opened 2017-02-11
Show locations on location index with image thumbnails, alphabetically


shacker:

heymonkeyriot:
@shacker are we talking about the location index page here?

e.g. the location index page needs to list all location pages alphabetically and with the thumbnail image of that location?

shacker:
Yes, sorry - was referring only to the Locations index here.

heymonkeyriot:
No worries. I've amended the title to make it a little clearer.

I'll work on the design bit of it today, and try and get it all in a PR for you to review when your back online PST time.

Issue #9, opened 2017-02-10
Complete styling work


shacker:

heymonkeyriot:
To mention I've started work on this in a separate branch. When I've gotten a little further I'll push up a PR to work on breaking down the different tasks I think we need to complete.

daaray:
Assuming that since the PR is merged, we can deal with other items in followup tickets.

PR #32, opened 2017-02-15
Location styling

### LINKED ISSUE #31: Location images being stretched to 100% on index page
With:

`{% image location.image width-150 %}`

we get this:

screenshot 2017-02-15 00 28 57

heymonkeyriot:
Should close #31, #24, and works towards #9

 - [x] Two col location index page
    - [x] Title
    - [x] Thumbnail image
    - [x] Address
    - [x] Link to map
    - ~~[ ] If location is open~~
 - [x] Location detail page
    - [x] Image
    - [x] Title
    - [x] Opening hours
    - [x] Map
    - [x] Fixing mouse scroll Google map issue (currently we have a brittle pointer-event CSS hack that I introduced)
    - [x] Introduction
    - [x] Body
    - ~~[ ] Highlight opening times for today() (in a timezone naive way)~~

 - [x] Fix merge conflicts
Apologies for poor branch name.


shacker:
By "mouse scroll issue" do you mean what happens when the scroll goes over the map and switches to map zoom? That could be mostly solved by not running the map full-width, so you can scroll over whitespace. We don't want to lose the ability to zoom the map.

heymonkeyriot:
Yup, exactly that. Though @shacker you'd still have the issue that if the mouse cursor is over the map your scroll will be hijacked, even if the map's smaller it doesn't really solve the issue, just makes it slightly less likely to happen. We definitely don't want to lose the functionality of users being able to click/ zoom in the map though. It's a reasonable well known problem, and the solution is to just swap out a class with JS if the user clicks to the map.

heymonkeyriot:
Think this is code complete. I'm going to deal with the time parts of this on a future PR since they're proving slightly more time consuming that I expected to do elegantly.

Note that it'd be good to do a second pass on this design when the other sections of the site are styled, but I think it's okay for now.
screencapture-localhost-8000-location-index-page-1487196837008
screencapture-localhost-8000-location-index-page-location-1487196676466

daaray:
There are some conflicts that need to be resolved. The styling on the branch is great; uncertain how the new code from master should be (or not be) incorporated.

🚀

heymonkeyriot:
I'll take on the merge conflicts. I think it's probably easiest for me to address them.

heymonkeyriot:
@daaray I've fixed the merge conflicts. @shacker to mention I think yesterday you committed a couple of things that conflicted with this branch direct to master. Would be good if you can review too before it gets merged. I'd suggest that we deal with any of those issues as a subsequent issue/ PR (e.g. whether to show the location index page image on the template or not)

bakerydemo/templates/locations/locations_index_page.html
```diff
@@ -1,18 +1,36 @@
 {% extends "base.html" %}
-{% load wagtailcore_tags %}
-{% load wagtailimages_tags %}
+{% load wagtailcore_tags navigation_tags wagtailimages_tags %}
 
 {% block content-header %}
-    {% include "base/include/header.html" %}
```
shacker: @heymonkeyriot I had gone through and moved all of the headers into a common header include. It looks like that has been lost here. Intentional? If not, can we preserve or refactor that system? (OK if we do that after this lands).

shacker:
I suspect this will also result in conflicts with the migration files in PR #47 , but at this point we should just go with whichever is completed first, then the other can manually resolve those.

shacker:
I've tested and merged this to keep the ball rolling, but it raised a few questions (I'll make separate tickets):

- Restore global header include
- Styling for me doesn't look for me like it does in the screenshot above?
- Google map code is still in place, but on-page maps are broken. Do we want both on- and off-page maps?

Issue #27, opened 2017-02-12
Add documentation and sample settings to support S3 on Heroku


daaray:
Heroku's ephemeral file storage is problematic for demos that last more than a few hours.  While not free, S3 support via `django-storages` would allow for a very low cost persistent solution.

daaray:
_This may be my lack of knowledge with Heroku and S3, but presumably the only way to implement S3 on Heroku easily is with the Heroku bucketeer?_

@heymonkeyriot bucketeer looks like it would resolve this issue; we could write documentation for how to do this 'by hand' and then the user would only be charged for actual S3 costs (pennies instead of $5 a month).

I am in favor of authoring the supplemental docs to DIY the config and not advocate for a particular heroku addon.

daaray:
We can repurpose content from https://wagtail.io/blog/amazon-s3-for-media-files/ or https://www.caktusgroup.com/blog/2014/11/10/Using-Amazon-S3-to-store-your-Django-sites-static-and-media-files/

PR #34, opened 2017-02-15
Pin versions of requirements

### LINKED ISSUE #28: Pin requirements
The requirements file should be amended to pin the current working versions of all libraries (at the time this demo is being actively developed). Leaving it to the most recent versions will invariably cause a problem later on.

daaray:
Fixes #28 

Over time, if we do not pin the requirements, some dependency will invariably get out of sync and cause problems for the potential user.  This PR pins to the latest known working version of Wagtail and the underlying dependencies.


requirements/heroku.txt
```diff
@@ -1,6 +1,6 @@
 -r base.txt
 # Additional dependencies for Heroku deployment
-dj-database-url
-gunicorn
-psycopg2
-whitenoise
+dj-database-url==0.4.1
+gunicorn==19.6.0
+psycopg2==2.6.2
+whitenoise==3.2.2
```
shacker: Does the presence of whitenoise here mean Heroku users will not be serving media files via apache or nginx? That's very cool actually.

daaray: [whitenoise](http://whitenoise.evans.io/en/stable/) resolves serving the static assets. Uploaded media is still handicapped by Heroku's ephemeral filesystem, hence the discussion taking place in #27

PR #45, opened 2017-02-18
Add Dockerfile and docker-compose.yml

tobiasmcnulty:
* Add Alpine Linux-based Dockerfile with optimizations to minimize the image size
* Switch to uWSGI (can be configured entirely via environment variables) and includes a built-in HTTP proxy which survives better "in the wild" (see http://uwsgi-docs.readthedocs.io/en/latest/HTTP.html#can-i-use-uwsgi-s-http-capabilities-in-production)
* Refactor the Heroku-specific requirements and settings files into "production" files; the latter uses environment variables to keep sane defaults but allow customization

To do:
- [x] Add defaults for S3 uploaded media (per `readme.md`) to production settings file
- [x] Add `.travis.yml` for testing Docker image build (can be expanded later to run unit tests, if any)


daaray:
Two things:

1. https://github.com/tobiasmcnulty/bakerydemo/blob/docker/app.json#L7 needs to be changed to `    "DJANGO_SETTINGS_MODULE": "bakerydemo.settings.production"`

2. I cloned this demo, checked out the docker branch, and ran `docker-compose up --build`.  I get the following error, any thoughts?

screen shot 2017-02-17 at 9 39 07 pm

daaray:
- Confirming latest edits resolved my issues
- Confirming deploy to heroku button build works with this  PR

tobiasmcnulty:
Thanks! I'll get S3 in there and then it should be ready to merge. Do you
know if there are plans to add Travis (or another CI)? We could also
include a smoke test for the docker image.

daaray:
We definitely did not get so far as to have a discussion regarding CI, but I think that it is an excellent idea, given that we have multiple contributors.

tobiasmcnulty:
@daaray I could not get S3 working with boto3, but this configuration seems to work. Might have been because I'm relying on `AWS_AUTO_CREATE_BUCKET`. Did boto3 work for you?

daaray:
Yes, I did utilize boto3, per the blog post I derived the material from: https://wagtail.io/blog/amazon-s3-for-media-files/

tobiasmcnulty:
This appears to be the issue: https://github.com/jschneier/django-storages/issues/268

Given this and the number of other `S3Boto3Storage` issues recently resolved and/or still outstanding in django-storages, I suggest we stick with boto2 for now.

daaray:
Tested latest (pre Travis commits):

- Docker build 
- Heroku w/o S3
- Heroku w S3

🚢 when you are happy with the Travis config.

raphaa:
Hey there! I just came across this issue and experienced the same killed by signal 11 when using django-storages in Docker. I've used python:3.6-alpine and then retried it with python:3.6-jessie and suddenly it works. Don't know why, but better to use boto3 than boto. Just fyi :)

kaedroho:
Alpine linux doesn't have glibc so the Python implementation is probably linked against musl instead. I guess that has something to do with the crash you're getting. Seems a lot of Python projects run into an issue with musl's default stack size (which is only 80kb).

gasman:
Yep, we've seen issues with Alpine Linux before: https://github.com/wagtail/wagtail/issues/3438. The conclusion I came to then was that the default system config was too resource-constrained for us to sensibly support.

PR #48, opened 2017-02-18
Logic to determine if a store is currently open

### LINKED ISSUE #38: Create templatetag that determines if a location is open or closed
In a timezone naive way, we should be able to use the opening hours model you created to show whether a location is open. It'd be a further nice way to show the building blocks

daaray:
Closes #38 

@heymonkeyriot: After diving in, I realized a templatetag is a bit too heavy handed and not necessary to determine if a location is currently open.  

Instead, I added a model method.  Still useful, IMO.

### DIFF
--- a/bakerydemo/locations/models.py
+++ b/bakerydemo/locations/models.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+
+from django.conf import settings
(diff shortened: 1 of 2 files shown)


shacker:
LGTM.

PR #53, opened 2017-02-19
Amend location page to include styling

heymonkeyriot:
Including style amending the tags (to `page.operating_hours`) and html tags.

### DIFF
--- a/bakerydemo/templates/locations/location_page.html
+++ b/bakerydemo/templates/locations/location_page.html
@@ -7,11 +7,6 @@
         #map {
           height: 100%;
         }
-        html, body {
-          height: 100%;
-          margin: 0;
-          padding: 0;
-        }
     </style>
 {% endblock head-extra %}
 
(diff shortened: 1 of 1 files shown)


heymonkeyriot:
@shacker would you be able to check that this branch is styled as per https://github.com/wagtail/bakerydemo/pull/32 before it's merged to master?

shacker:
When I check out this branch, map works great - doesn't trap mouse scroll, but still interactive. The display is as shown here, and does not match #32.

screenshot 2017-02-19 13 46 45

heymonkeyriot:
Have you included an introduction when you populated the page?

shacker:
Ah yes, that's much nicer. I'll get intros into all of the locations. 

That still leaves the styling of the hours at bottom left of the map, like this (all the way up against  the left edge of the viewport):

screenshot 2017-02-19 14 03 15

heymonkeyriot:
Those lines from the template shouldn't be there

At line 83 there was
```
    {% if page.is_open %}
      Open Now!
    {% else %}
      Currently Closed
    {% endif %}

    {% for hours in page.operating_hours %}
        <li>{{ hours }}</li>
    {% endfor %}
```

but that is deleted in commit 0d5ce5a

The page should (minus the odd black line that the screen capture tool added) look like
screencapture-localhost-8000-location-index-page-location-1487196676466

shacker:
Thank you!

PR #55, opened 2017-02-19
Refactor WagtailBakery admin menu item(s)

### LINKED ISSUE #22: WagtailBakery admin menu item
We are surfacing the People snippet under WagtailBakery. But People and the other three snippet models are already present under Snippets. Do we want to move all snippets under WagtailBakery? Or remove that and leave them all under Snippets? (I vote for the latter - it's more standard).

shacker:
- Access to People, FooterText, BreadType and Country from custom menu items
- Now uses wagtail_hooks  rather than putting config in the models
- Resolves #22

### DIFF
--- a/bakerydemo/base/models.py
+++ b/bakerydemo/base/models.py
@@ -5,8 +5,6 @@ from django.db import models
 from modelcluster.fields import ParentalKey
 from modelcluster.models import ClusterableModel
 
-from wagtail.contrib.modeladmin.options import (
-    ModelAdmin, ModelAdminGroup, modeladmin_register)
 from wagtail.wagtailadmin.edit_handlers import (
     FieldPanel, FieldRowPanel, InlinePanel, MultiFieldPanel,
     PageChooserPanel, StreamFieldPanel,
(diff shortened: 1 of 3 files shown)


shacker:
Still not sure this is optimal. Thoughts? Now appears as:

screenshot 2017-02-19 15 22 14

screenshot 2017-02-19 15 22 23

heymonkeyriot:
It's definitely an improvement. We have font awesome installed for icons, so could maybe use that to help make it visually clearer that these elements are different to `Explorer`? I'm afraid I've no idea how to improve "Wagtail Bakery" but it definitely feels like a slightly title for the section.

shacker:
Updated with external font-awesome icons, resolved merge conflict.

shacker:
I wish we could nest the custom menu items down more than one level, but unfortunately it doesn't seem to be possible. Added doc note on both the bundled FA icon set and us using the full set via CDN.

daaray:
🌟 🌟 🌟 🌟 🌟

PR #61, opened 2017-02-21
Use blog intro field on blog index, detail; blog bylines, avatars

shacker:
- Resolves #60

PR #63, opened 2017-02-22
Blog index styling

heymonkeyriot:
**READY FOR REVIEW**

Note: I've merged off of #61 on the hope that that will reduce the number merge conflicts (will obviously deal with any merge conflicts if they do come up)

Styling is based loosely on a simplified version of [Torchbox's blog](https://torchbox.com/blog/) for lack of my own imagination

 - [x] Style tags
 - [x] Style columns
 - [x] Style blog items
 - [x] Try and reduce the amount of verbosity I'm using in the CSS


heymonkeyriot:
Slight issue here that we don't currently - that I can see - get a list of tags to return outside of the individual blog object on the index page. The loop that is currently in the template doesn't return anything, and I can't see how it would do. @daaray am I being really daft and missing something? Happy to make the changes to the BlogIndexPage model if the isn't already there but might do as a different PR since it feels quite different from the styling.

heymonkeyriot:
So I think this is ready for review.

Various different viewports
screencapture-localhost-8000-oi-1487888175931
screencapture-localhost-8000-oi-1487888216901
screencapture-localhost-8000-oi-1487888239621
screencapture-localhost-8000-oi-tags-tag2-1487888356719

**Some questions:**
I've rather stubbornly not put the images as background elements in divs. That would be easier to style, but is semantically wrong and would mean we couldn't use `<picture>` elements. Since we're not using `<picture>` elements anywhere though I'm totally open to someone suggesting that we should just make them background images as a slightly more pragmatic way of dealing with them. Thoughts?

I've not dealt with edge cases where there are an uneven number of blog items. That feels like a nice to have, and something I'd be happy to wheel back to, but not something I feel needs to be dealt with in this PR?

I've used flexbox, and though I haven't tested, I doubt this will play well with a non-flexbox browser. I don't think that's necessary. Looking at [caniuse](http://caniuse.com/#feat=flexbox) it has 86% coverage. Do you guys think that's okay?

Probably something for an issue/ other PR but tags feel a little sub-optimal at the moment insomuch as we lose reference to all the other tags and lose reference to the index page title and intro. I wonder whether we should call a specific tag template page that's slightly simpler than the design blog one?

bakerydemo/blog/models.py
```diff
@@ -138,10 +138,29 @@ def tag_archive(self, request, tag=None):
...
-        blogs = BlogPage.objects.filter(tags=tag).live().descendant_of(self)
+        blogs = self.blogs(tag=tag)
 
         context = {
             'tag': tag,
             'blogs': blogs
         }
         return render(request, 'blog/blog_index_page.html', context)
+
+    def blogs(self, tag=None):
```
shacker: Not a blocker, but I would probably call this `get_posts()` (preferable to use `get_` for anything that retrieves other data, and distinguishing between a blog, which is the whole thing, vs a post, i.e. an individual post in a blog).

bakerydemo/blog/models.py
```diff
@@ -138,10 +138,29 @@ def tag_archive(self, request, tag=None):
...
+        blogs = BlogPage.objects.live().descendant_of(self)
+        if tag:
+            blogs = blogs.filter(tags=tag)
+        return blogs
+
+    def get_child_tags(self):
+        """
+        Returns the list of Tags for child pages.
+        """
+        tags = [x.get_tags for x in self.blogs()]
```
shacker: This is a nice feature! But I think this might be one of those areas where dense list comprehensions save a bit of code but make the code less easy to understand. I might use for loops here, or a for loop combined with list comprehension. I'll merge this and take a crack at it.

bakerydemo/templates/blog/blog_index_page.html
```diff
@@ -1,22 +1,55 @@
...
+            {% for blog in blogs %}
+                <li class="col-xs-12 col-sm-6 col-md-3 blog-list-item">
+                <a href="{% pageurl blog %}">
+                    <div class="image">
+                        {% image blog.image fill-850x450-c50 as image %}
+                        <img src="{{ image.url }}" width="{{ image.width }}" height="{{ image.height }}" alt="{{ image.alt }}" class="" />
+                    </div>
+                    <div class="text">
+                        <h2 class="blog-list-title">{{ blog.title }}</h2>
+                        <p>{{ blog.introduction|truncatewords:15 }}</p>
```
shacker: Isn't the point of having an introduction field so that we don't have to do a truncation? If truncating to make the design work, maybe better to limit size of the introduction field itself. I can see it both ways. Not a blocker.

shacker:
+1 on Flexbox.

I have an uneven number of posts and it doesn't look weird - just stops where it stops. All good. 

> tags feel a little sub-optimal at the moment insomuch as we lose reference to all the other tags and lose reference to the index page title and intro.

Let's feel it out and do a separate PR if needed.

shacker:
Since you branched off #61, we should be able to close that one without merging if this one is merged right?

bakerydemo/templates/blog/blog_index_page.html
heymonkeyriot: Normally on a production site where we show introductions both on the listview and detailview we have two separate fields for this reason, since the introduction on the detailview can be substantially longer than on the listview. Personally I think here it's okay to just have the single field with the check on the blog index page, though totally open to other solutions.

heymonkeyriot:
Yes, exactly right @shacker. #61 can be closed. The comment I left on it (e.g. whether we want to have the square image etc.) can be picked up later in a different issue/PR.

PR #67, opened 2017-02-23
Place sqlite file in conventional PROJECT_ROOT path

### LINKED ISSUE #37: referenced files that don't exist
readme says:

`cp bakerydemo/settings/local.example.py bakerydemo/settings/local.py`

bash says:

`cp: bakerydemo/settings/local.example.py: No such file or directory`

and:

`./manage.py runserver` yields:

```
    raise ImproperlyConfigured("The SECRET_KEY setting must not be empty.")
django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.
```

But secret key is already there in base.py. Is base.py no longer being read? Looks like `manage.py` now references:

`os.environ.setdefault("DJANGO_SETTINGS_MODULE", "bakerydemo.settings")`

But there is no `settings.py` in bakerydemo.  So I'm totally blocked and can't even runserver. Can't work on the project tonight. Grumble.

daaray:
Fixes #37: sqlite db file should be in the root directory of the entire project:

`/path/to/bakerydemo`

NOT

`/path/to/bakerydemo/bakerydemo`

Once merged, you'll need to move your current DB up one level else a new one will be created.

### DIFF
--- a/bakerydemo/settings/base.py
+++ b/bakerydemo/settings/base.py
@@ -13,8 +13,8 @@ https://docs.djangoproject.com/en/1.10/ref/settings/
 import os
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-
+BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir))
+PROJECT_ROOT = os.path.abspath(os.path.join(BASE_DIR, os.pardir))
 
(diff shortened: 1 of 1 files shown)


shacker:
Thanks for this, but aren't we moving to postgres? Or do we want to keep both RDBMSs in play?

daaray:
@shacker long term, I think it would be appropriate to move toward postgres. Short term, I think we should remediate the atypical location of the db file 🍨

shacker:
Thanks!

PR #68, opened 2017-02-23
Add helper method to get all BlogIndexPage's child BlogPage's tags

daaray:
@heymonkeyriot here is your helper method. Apologies for the oversight!


heymonkeyriot:
Perfect thanks @daaray. Working perfectly so will merge in.

Issue #57, opened 2017-02-20
Styling Bread app


heymonkeyriot:

heymonkeyriot:
Pull request #71 has been merged into master so closing this issue.

PR #70, opened 2017-02-24
BreadIngredient <--> Bread M2M example

### LINKED ISSUE #64: Demonstrate ManyToMany
Wagtail 1.9 now supports true ManyToMany relations. This is a pretty big deal and erases a long-standing limitation. I know it's late in the game but can we think of a way to demonstrate this important feature?

daaray:
Fixes #64

 - add M2M example for bread model


heymonkeyriot:
This looks good. Two questions:

 - When there are no ingredients the user just sees an empty div, and it looks a little broken (even though it obviously isn't) I guess the only way around that might be some help text to tell people how to add an ingredient? Or is there something slightly more elegant we could do. Even with the ForeignKey models for 'origin' and 'bread type' there's at least an empty dropdown.
 - It may not be something for the moment, but I wonder whether we could augment the `BreadIngredient` model with how much of each is required? e.g. 350g flour, rather than just 'flour'?

daaray:
To be honest, while implementing it, I became concerned that this might not be the best example for a M2M in Wagtail.  From the [docs](http://docs.wagtail.io/en/v1.9/getting_started/tutorial.html?highlight=ParentalMany#categories), it seems as though a better use case for creating relationships where the related content is a fairly controlled, small list of things.

Ingredients can balloon out very quickly. What if we changed from `BreadIngredient` to `DietaryRestriction` and included fixtures for:

- Vegan
- Paleo
- Kosher

Any others?

heymonkeyriot:
That could work. Perhaps also

- Vegetarian (to distinguish that a bread doesn't use lard, but does have honey or egg etc, which wouldn't be vegan)
- Gluten Free
- Halal

(I may be wrong, but I think if we have vegetarian and vegan then we cover both Halal and Kosher as requirements. So long as bread is parve (e.g. with no animal product) it's kosher, and I think Halal follows the same principal)

shacker:
I personally feel this is ready to merge (though I would want to display the ingredients either as comma separated or as a standard list (not display-inline)). 

Re: The div appearing empty in admin: It won't for demo site users since we'll be pre-populating it with maybe a dozen ingredients. Actually, what we could do here is demo the "collapsed" property in the admin so it doesn't take up a lot of space. 

I don't see a need to switch from ingredients to kosher/vegan etc. - I think ingredients is a natural fit. 

Re: the list size balooning when lots of ingredients are added - This is indeed a limitation of wagtail. Django supports the raw_id field and [filter_horizontal](https://docs.djangoproject.com/en/dev/ref/contrib/admin/#django.contrib.admin.ModelAdmin.filter_horizontal), both of which I feel Wagtail needs to support, but that's not a problem we can solve here. All we need is to prepackage a dozen ingredients to get them started with the feature.

I think adding ingredient amounts (350g) is overkill for this purpose.

daaray:
@shacker thanks for the tip on the collapsible panel.  I think that is a good solution for my concerns.

heymonkeyriot:
All sounds good. I'll make sure to merge this in to the #57 feature branch and style as per comment from @shacker.

PR #71, opened 2017-02-24
Bread styling

heymonkeyriot:
**Ready for review**

 - [x] Style index page
 - [x] Amend context on index page model to pass through all fields
 - [x] Style detail page


bakerydemo/breads/models.py
```diff
@@ -94,26 +94,27 @@ class BreadsIndexPage(BasePageFieldsMixin, Page):
...
-
-        paginator = Paginator(all_resources, 5)  # Show 5 resources per page
+    @property
+    def breads(self):
+        return BreadPage.objects.live().descendant_of(
+            self).order_by('-first_published_at')
 
+    def paginate(self, request, *args):
         page = request.GET.get('page')
+        paginator = Paginator(self.breads, 2)
```
heymonkeyriot: I've put `2` just for the moment to make testing easier so you only need to create three bread objects to test pagination.

daaray:
Looking good; a few comments:
- We need a margin between the entries when the window is narrowed and the entries stack
- images should have a fixed width (or become hidden?) when the window narrows else they get distorted.

screen shot 2017-02-25 at 4 19 03 am

bakerydemo/breads/models.py
```diff
@@ -94,26 +94,27 @@ class BreadsIndexPage(BasePageFieldsMixin, Page):
...
-    def get_context(self, request):
-        context = super(BreadsIndexPage, self).get_context(request)
-
-        # Get the full unpaginated listing of resource pages as a queryset -
-        # replace this with your own query as appropriate
-        all_resources = self.get_children().live()
-
-        paginator = Paginator(all_resources, 5)  # Show 5 resources per page
+    @property
+    def breads(self):
```
daaray: To conform to convention, as @shacker pointed out in the Blog model, this should be renamed `get_breads` and not be a property.

heymonkeyriot: No problem to change the name to `get_breads`. But removing the `@property` raises an error `object of type 'method' has no len()` on the pagintor. I'm not sure why that's happening. Would welcome suggestions on how to resolve.

daaray:
@heymonkeyriot I pushed a change for; it looks as though when you changed from the property to a method, you weren't invoking the method:

`self.get_breads` vs `self.get_breads()`

heymonkeyriot:
`:face-palm:` knew it'd be something daft! Thanks @daaray.

heymonkeyriot:
Screenshots for desktop and mobile for the BreadIndexPage and BreadPage
screencapture-localhost-8000-breads-index-page-1488049657437
screencapture-localhost-8000-breads-index-page-1488133549236
screencapture-localhost-8000-breads-index-page-afoioingefffega-1488133405803
screencapture-localhost-8000-breads-index-page-afoioingefffega-1488133427368

bakerydemo/templates/breads/bread_page.html
```diff
@@ -10,7 +10,7 @@
                     <h1>{{ page.title }}</h1>
                     
                     <figure class="hidden-md-up">
-                        {% image self.image width-600 %}
+                        {% image self.image width-500 %}
```
heymonkeyriot: This means we only need to load a single image asset for the mobile viewport and desktop, which makes me feel a little better about using the visibility class.

bakerydemo/templates/breads/bread_page.html
```diff
@@ -2,24 +2,131 @@
...
+                </div>
+            </div>
+
+            <div class="col-md-5 col-md-offset-1">
+                <div class="row">
+                    <figure class="hidden-md-down">
+                        {% image self.image width-500 %}
+                    </figure>
+                    <ul class="bread-meta">
+                        {% if page.origin %}
```
shacker: I think it's slightly confusing to use self.image and page.origin in the same template. Would these both work with `page.foo` ?

shacker:
This is looking great.

Side question (not quite about this PR): I notice you put some thought into punny bread titles - they're very good. I'm already working with Tumi's original breads and have them in the "master" content copy. Were you wanting to go with funny title instead, or was that just placeholder content?

shacker:
Travis is not happy with Docker.

bakerydemo/templates/breads/bread_page.html
heymonkeyriot: Good spot, and absolutely correct that we should keep them consistent. Will amend to `page.foo` throughout and push up.

heymonkeyriot:
> Side question (not quite about this PR): I notice you put some thought into punny bread titles - they're very good.

I was sort of doing them for my own amusement, and was just placeholder content. I'm not sure how well they'll travel. I'd suggest we keep it fairly neutral in tone, though perhaps it wouldn't hurt to hide one or two puns around the place?

shacker:
Roger dodger, we'll stick with wikipedia content with maybe a pun here or there. 

I had parents in town this weekend and wasn't able to put in time, but back on the case tonight hopefully.

Issue #15, opened 2017-02-10
Contact Form styling


shacker:

shacker:
A demo contact form with these fields has been created and will be in the full data dump. 

Subject
Your Name
Your Email
Purpose [dropdown choices = Comment, Question, Feedback]
Body [textarea]

`EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'`

is already set in base.py and heroku.py to ensure the form does not send live email, and this is noted in the demo form and readme.

Re-assigning to @heymonkeyriot for styling.

heymonkeyriot:
Thanks @shacker 

From doing a very quick play earlier this morning I think Bootstrap does most of the heavy lifting with this, just need to make sure that all the field items are semantically marked up

shacker:
I can take this over if you're too busy. Looks like we'll need to switch from `form.as_p` to rendering the form fields manually in the template to put the correct styling on them.

heymonkeyriot:
Oh, sorry @shacker I missed your comment on here. I'm going to start on this today so that we can get it running; conscious I let it slip slightly.

heymonkeyriot:
I think this can be closed right @shacker? Unless you'd like to keep it open for the work you're doing on the dummy data?

shacker:
Yeppers.

PR #73, opened 2017-02-27
Styling form page

heymonkeyriot:
**Ready for review** see [comment below](https://github.com/wagtail/bakerydemo/pull/73#issuecomment-282970038) for more details

c/f #15 for original issue

 - [x] Use `{% for field in form %}` rather than `{{ form.as_p }}` in template
 - [x] Style form detail page
 - [x] Style form "thank you" page
 - [x] Fix merge conflicts
 - [x] Understand why date field and datetime field are rendering as type=text
 - [ ] Creating template override for date field and datetime field as per comment from @daaray


heymonkeyriot:
@daaray and @shacker, if you have the time or inclination to have a look at what might be causing the date field and datetime field to render with a form input `type=text` that'd be awesome. It looks very much as though it's an issue with the AbstractFormField, though having had a quick look at [forms.py](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailforms/forms.py) and [models.py](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailforms/models.py) of the forms app in Wagtail I can't see any obvious reason why these fields would fail to be correctly cast.

It's simple to replicate the issue, just add either a date field or datetime field to the page and see that it renders as the incorrect field type.

daaray:
@heymonkeyriot we'll have to override the templates, I believe: https://github.com/wagtail/wagtail/issues/1016

heymonkeyriot:
Thanks @daaray. Having done a very limited amount of digging I wonder whether it might be better to do the overriding work as a separate PR/ issue. I think the PR, taking it at it's strictest definition of being about styling the form page, is complete, and think it might be worth merging in. What do you @shacker or @ArnarTumi think?

daaray:
I agree that it should be dealt with as a separate issue.

shacker:
Since we don't have a date/time field on the contact form, that issue doesn't really affect us (or do we need one? Purpose/intent?). 

I personally would rather see wider fields on the desktop view, but that can be solved separately. 

Thanks!

shacker:
We seem to have gotten into a migration quandary, with two `base.0013` files. I did a migration merge which references files

```
    dependencies = [
        ('base', '0013_auto_20170227_2216'),
        ('base', '0013_auto_20170228_0714'),
    ]
```

but I only have 0013_auto_20170228_0714, not the other. Did it get checked in?

heymonkeyriot:
Yup commit  [600bfc4 ](https://github.com/wagtail/bakerydemo/pull/73/commits/600bfc482df833db9dc18fc48d76290f8657b72d0) on this branch pushed up `0013_auto_20170227_2216.py`. I can see it on the [master tree too](https://github.com/wagtail/bakerydemo/tree/master/bakerydemo/base/migrations)

It seems the conflict has come from the removal of the location relationship on the other branch.

shacker:
Huh - not sure what happened there, but all resolved now. I didn't have the latest somehow, and tried to create the missing merge migration, which failed... Reset just now, started over, got it all resolved and pushed the missing migration merge. Sorry for the noise.

Issue #75, opened 2017-02-27
Styling about page


heymonkeyriot:
Not really sure it can be described as styling, but needs the grid around the body.

Issue #54, opened 2017-02-19
Handle locations on About template, or ditch that relationship


shacker:
Locations are shown as children of `LocationIndex`. About also has `AboutLocationRelationship`, which lets you manually attach those same locations to the about page. But if you do, they don't show up since the template doesn't handle them. 

Either add that handling to the about template, or remove the `AboutLocationRelationship` since it duplicates data on Locations and we're going to use About for info about  the demo itself anyway.

heymonkeyriot:
I think we should drop the relationship. We show in the demo how this type of demo can be set up between `People` in the base app, and `BlogPage` in the blog app. I think given the copy you've shared on #4 works well, and having the location relationship there will just confuse the issue. If we change our mind we can always put it back in again.

PR #76, opened 2017-02-28
About template styling (and amended header include)

heymonkeyriot:
c/f issue #75

Adding bootstrap `<div>`s to about page, and amending the generic header include that has the image with an auto width on desktop (with max-width of 100% of the column), whilst on mobile it's 100% as below. I haven't included the location relationship on the template as I'm not sure it fits the copy that @shacker has been working on.

screencapture-localhost-8000-about-page-1488264873918
screencapture-localhost-8000-about-page-1488264889681

One thought I had with this, especially if we remove the location relationship as per #54, is whether we should describe this as an "about page". On most of the Wagtail sites I've worked on we have some sort of standard page that have no special attributes but can be used to create these sort of pages (be they about pages, privacy policies etc.) Would we be better off changing the name of this model?


shacker:
Nice, thanks!

PR #77, opened 2017-02-28
Remove  aboutlocations

### LINKED ISSUE #54: Handle locations on About template, or ditch that relationship

shacker:
- Overlaps with functionality of Locations view
- Resolves #54


daaray:
👍

heymonkeyriot:
Looks good testing here.

daaray:
oops I flew in under the radar. Sorry @heymonkeyriot.

heymonkeyriot:
No worries at all :) It's always good to have two pairs on eyes

Issue #58, opened 2017-02-20
Homepage styling and content


heymonkeyriot:
I think the issue between styling of the homepage and the content it has is quite interlinked. We didn't, in the sprint, have time to really have a conversation about what should be on the homepage. A little like the about page it'll need to juggle - in my mind - the content model of the bakery chain, alongside being clear it's a demo for Wagtail.

Alongside any homepage fields we have the following apps we can show content from:

 - About page
 - Blog page
 - Bread page
 - Locations
 - Gallery

I wonder whether we could have a content model similar to a shop/ chain, though the first screen the user sees is relevant to Wagtail rather than the bakery? For the sort of creative positioning I had in mind: 

 - http://www.traderjoes.com/
 - http://www.planetorganic.com/
 - https://www.riverford.co.uk/

Happy to draw something up if useful? But was also curious to hear what others had thought.

shacker:
My personal vote would be to keep it short and simple, probably just a single StreamField with no data integrations. Willing to be outvoted on this.

heymonkeyriot:
My concern about it being a StreamField element is that it might be a little difficult to style. If we're going to keep it simple I'd suggest we at least keep it field based. I'll try and think of something a little more concrete to propose than my admittedly vague opening post above.

shacker:
Classic title and either RichTextField or TextField body would be fine too.

shacker:
For reference: Ben's original design: 
http://playground.torchboxapps.com/bakery-html-proto/standard.html

heymonkeyriot:
That was actually Ben's design for the blog page/ streamfield elements. The design for the homepage (which wasn't really completed, and based on the proposal from @tomdyson) is at http://playground.torchboxapps.com/bakery-html-proto/

shacker:
Ah! I missed that somehow, seeing now for the first time. Good timing, I was about to load up dummy homepage content. What do you think - do you want to keep this "departure" look and feel? Or stick with a streamfield with dummy content for the hp? I think we definitely want a hero image for it. If we keep Ben's design, what's the best way to model - a new HomePageSection model for the subheads and their texts?

heymonkeyriot:
I'd like to stick closer to the proposal from @tomdyson and @benenright for sure. I think trying to keep it within a StreamField will be quite limiting, and won't have that eye-catching aspect that I think we're hoping for.

[...]

PR #80, opened 2017-02-28
Homepage style

heymonkeyriot:
**Not ready for review**

 - [x] Amend HomePage model
 - [x] Add functions to IndexPage models to access child items
 - [x] Template formatting
 - [x] Template styling


heymonkeyriot:
I didn't get quite as far with this as I'd hoped but thought it was worth pushing where I'd gotten to. Currently looks like this:
screencapture-localhost-8000-home-page-1488321108571

I'm unsure about the positioning of the StreamField element. It perhaps should be after the promo block and the first group of featured articles?

heymonkeyriot:
I think this is ready to review and/or have a conversation about. I think there will need to be tweaks once @shacker you've had a chance to decide on content. Below screenshots of desktop and mobile. There's still some awkward breakpoints, but I think these can be refined once someone other than me has had a chance to look at this :)

screencapture-localhost-8000-home-page-1488406859355
screencapture-localhost-8000-home-page-1488407840075

bakerydemo/base/models.py
```diff
@@ -159,14 +159,126 @@ class HomePage(Page):
         related_name='+',
         help_text='Homepage image'
     )
+    hero_text = models.CharField(
+        max_length=255,
+        help_text='Write an introduction for the bakery'
+        )
+    hero_cta = models.CharField(
+        verbose_name='Hero CTA',
```
shacker: I don't know what CTA means - will user-devs?

bakerydemo/base/models.py
```diff
@@ -159,14 +159,126 @@ class HomePage(Page):
...
+    )
+    featured_section_3 = models.ForeignKey(
+        'wagtailcore.Page',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+        help_text='Third featured section for the homepage. Will display up to six child items.',
+        verbose_name='Featured section 3'
     )
```
shacker: Corresponding to your screenshots, are these 3 "featured sections" the blog posts? If the goal of this HP design is to feature a bit from each main section of the site, can't we just have a model method that pulls the last 3 blog posts? I think this technique of hard-coding 3 separate objects would not be regarded as a best practice.

bakerydemo/static/css/main.css
```diff
@@ -868,3 +1092,28 @@ span.outline {
...
+}
+.responsive-object iframe,
+.responsive-object object,
+.responsive-object embed {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+}
```
shacker: Minor, but seems like you have a mix of 2-space and 4-space indentation in CSS. Personal pref, but my vote is 4-space indentation in CSS just to keep things simple and consistent with python.

bakerydemo/templates/base/home_page.html
```diff
@@ -2,6 +2,125 @@
...
+                    {{ page.promo_text|safe }}
+                </div>
+            {% endif %}
+        </div>
+
+        <div class="col-sm-6 col-sm-offset-1 feature-1">
+            {% if page.featured_section_1 %}
+            <h2>{{ page.featured_section_1_title }}</h2>
+                <div class="featured-children">
+                    {% for childpage in page.featured_section_1.specific.children|slice:"4" %}
```
shacker: Nice! Didn't realize we could use slice in template syntax. Thanks for that tip.

bakerydemo/templates/base/home_page.html
```diff
@@ -2,6 +2,125 @@
...
+                                <figure>
+                                    {% image childpage.image width-380 %}
+                                </figure>
+                                <h3>{{childpage.title}}</h3>
+                            </a>
+                        </li>
+                    {% endfor %}
+                </div>
+            {% endif %}
+        </div>
```
shacker: These hard-coded feature sections are really not dry. Better done in a loop from a queryset, IMO.

shacker:
This is beautiful and does a great job of previewing the major aspects of the site, but I am a bit surprised because I thought you wanted to stick to the  simpler [original design](http://playground.torchboxapps.com/bakery-html-proto/)  and that you preferred some custom fields and no StreamField (per comments in #58) ? So this feels like quite a departure. Not necessarily opposed, but I have to say it feels a bit complex for a homepage in a demo site. I can be swayed though. 

Thoughts from others? @daaray ?

bakerydemo/base/models.py
heymonkeyriot: Ah, maybe that's an English-English expression? It stand for "call-to-action". Happy to spell it out more clearly in the verbose_name. Unless you think there might be a better, more generic English expression?

bakerydemo/static/css/main.css
heymonkeyriot: There might be a mix since I've been a little hurried. I feel like 2-space is the more common convention for CSS/ SASS, but fairly agnostic. I think it might make sense to amend within the CSS refactoring ticket in any case. There's quite a lot of repetition in that file.

bakerydemo/base/models.py
heymonkeyriot: Not quite. Corresponding to the screenshots, featured section 1 = bread section; featured section 2 = blogs and feature section 3 = locations.

bakerydemo/templates/base/home_page.html
heymonkeyriot: The specific children of the index pages are coming via a loop (though admitedly not a query set, it's using TreeBeard's API to get the direct children). We could be even drier by using the index pages title rather than a homepage specific one, but that limits us around tone of voice. 'Blogs', 'Breads', 'Locations' etc are a bit dull

bakerydemo/base/models.py
heymonkeyriot: We could do this on the model side using querysets, but I can't see an obvious way to give the user control over what content they want to feature on which bit of the homepage. I've had a few sites where different sections need to be shown on different parts at different times.

bakerydemo/templates/base/home_page.html
heymonkeyriot: I confess this is something I'm not terribly happy about. It's definitely putting some logic on the template, but I couldn't work out a model-based solution to the problem of having three different featured sections one of which has four objects, one that has three and one with six.

bakerydemo/base/models.py
shacker: Ah! Makes sense. Probably used in many U.S. design houses and I just haven't come across it in .edu. Yeah the verbose text should clarify just fine.

bakerydemo/base/models.py
shacker: Oh I see what you're doing with this now. The goal makes sense, but I do think we could solve it with quite a bit less code by using a HomePageRelatedSection, which would inherit from Orderable (I don't think we're demo'ing Orderable and we probably should) and a FK to Page. Then they could be in a single (possibly collapsed) admin section. Downside is that they would all use the same rendering code so you'd lose the ability to have your section 1 look quite different from the others, but maybe that's OK for this purpose?

bakerydemo/static/css/main.css
shacker: +1

bakerydemo/templates/base/home_page.html
shacker: If we go with the DRYer approach from my previous comment, this template might just slice the first six of each set. If it only finds three, that's still OK.

shacker:
Interested to hear comments from others on the question of simpler design proposed by Ben and Tom, or this richer but more complex approach.

daaray:
I am in favor of the richer home page.  It leaves a better first impression for the intended audience as to what is possible.

bakerydemo/base/models.py
daaray: We are using `Orderable` for the operating hours:

`class LocationOperatingHours(Orderable, OperatingHours):`

heymonkeyriot: I think that's the problem, to have the very different styling (and positioning) we need the verbosity. We _could_ use a single loop and have different styling using the CSS's `nth-child` but that would be more code, and more complicated I think.

We have orderable in a few places (the bread & bread ingredients M2M, and I think the author on blogs)

shacker:
OK, richer homepage it is then. Let's go with this HP sections approach as-is, and if one of us has another brilliant idea we can change it later. Docker is unhappy, but that can also be resolved later. @heymonkeyriot anything else you want to change before we bring this in? 

Huge amount of work on this, thanks.

heymonkeyriot:
Personally I'd like to delete the StreamField on it :) But also think it's something we can do at a later point. I have the sensation that we're slightly trying to feel out an elephant in the dark because we're working on little bits and pieces at a time. I suspect once we've got all those pieces together we'll have a better sense of how they can be improved and made to work seamlessly together. Which is a very long winded way of saying, yes let's merge it in 🎉

shacker:
@heymonkeyriot Out of curiosity, are you not a fan of StreamField in general, or just not for our purposes here, or... ? (I have mixed feelings about it in general).

Issue #78, opened 2017-02-28
Rename AboutPage to more generic title?


heymonkeyriot:
I asked this question on PR #76

> One thought I had with this, especially if we remove the location relationship as per #54, is whether we should describe this as an "about page". On most of the Wagtail sites I've worked on we have some sort of standard page that have no special attributes but can be used to create these sort of pages (be they about pages, privacy policies etc.) Would we be better off changing the name of this model to something like StandardPage?

shacker:
+1 I think that's a good idea. We don't want to imply that devs have to create a new model for every variant of standard-issue pages. I'll take this one tonight.

heymonkeyriot:
Awesome!

shacker:
Soon realized that we already have `base.BasePageFieldsMixin` which defines introduction, image, and body as `blocks.BaseStreamBlock`, which left me wondering whether we really need `StandardPage`, or whether we're better off sticking with custom models inheriting from `base.BasePageFieldsMixin`. Two different techniques for abstraction  - which is better? Mixins have better composability, but `StandardPage` is a common pattern. 

Thoughts?

shacker:
(actually I already moved `body` into BasePageFieldsMixin` - you guys won't see that yet.

shacker:
This turned interesting.

I made an executive decision and decided to create a StandardPage, definining body, image and introduction. All good.

Then I went to remove the `BasePageFieldsMixin` and make all the app.models that use it subclass StandardPage instead, since those were redundant.  Ran into a big migration problem. `makemigrations` wants me to define a default value for `standardpage_ptr`.  I can't make it null or optional since it's a subclass, not a FK relation. I can't specify 1 since I can't guarantee that exists. I can't use `StandardPage.objects.first()` since the migration making tool doesn't recognize the import. It seems the only option is to wipe the db and start over but don't want to do that. Stuck.

Now feeling it would be better to keep `BasePageFieldsMixin` and ditch StandarPage. Though I could wipe the db if anyone feels strongly about it one way or another.

Good object lesson in why mixins are more flexible than StandardPage-style approach,

shacker:
OK finally figured out how to do this cleanly. Notes at
https://github.com/wagtail/bakerydemo/pull/87

PR #87, opened 2017-03-03
Clean up relations between BasePageFieldsMixin, StandardPage

shacker:
This is a tricky one, responding to #78. Summary:

- We already had BasePageFieldsMixin defining image and introduction, which many of our models subclassed. 

- We redefined `body` on almost every model. Made sense to bring body into BasePageFieldsMixin instead.

- We wanted a new StandardPage model for "generic" pages. With the above in place, now able to make a new StandardPage model that just wraps BasePageFieldsMixin with no additional fields (the BasePageFieldsMixin gives us code sharing between models while the StandardPage offers this abstraction as a page type in the admin)

- Remove old AboutPage model

Tried several approaches and ran into unexpected dead ends, but finally got the right order of operations to make it work without destroying existing content.

A bit nervous that this will bork on current devs' installations - let me know if it does, though we may resort to bringing you back up with a fresh db.


daaray:
I am ok with a local dev having to reset their DB; after all, we are going to squash the migrations prior to release, yes?

However, on a fresh DB, running `migrate`, I get the following:

`django.db.migrations.exceptions.NodeNotFoundError: Migration base.0017_standardpage dependencies reference nonexistent parent node ('wagtailimages', '0019_delete_filter')`

We have seen this before, I think?  Are you using Wagtail [master](https://github.com/wagtail/wagtail/tree/master/wagtail/wagtailimages/migrations) and not [1.9](https://github.com/wagtail/wagtail/tree/stable/1.9.x/wagtail/wagtailimages/migrations)?

shacker:
@daaray I can't reproduce this problem. If I move my bakerydemodb out of the way and run migrate, the database builds cleanly. Are you sure that was a fresh db you were testing with?

heymonkeyriot:
Will try and test here this evening. I'm also okay squashing migrations given that we have an issue to cover that later. 

My concern is that it's only a small majority of pages that have body fields at the moment. Currently GalleryPage, Bread index page, Locations index page and Blog index page don't have body fields, and I don't think they should. The image for them works because it makes sense to have a listing image if they're being referenced by another page, the introduction works because we show it on their own template pages, but I don't think the same can be said for the body field. What do others think?

bakerydemo/base/migrations/0017_standardpage.py
```diff
@@ -0,0 +1,35 @@
...
+import wagtail.wagtailcore.fields
+import wagtail.wagtailembeds.blocks
+import wagtail.wagtailimages.blocks
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailcore', '0032_add_bulk_delete_page_permission'),
+        ('wagtailimages', '0019_delete_filter'),
```
daaray: @shacker I did use a clean instance with no DB. This line is the offender; the reference to 0019 leads me to believe that this migration was created in a virtualenv that is using Wagtail master branch, as the 1.9 branch does not have this migration.

shacker: Ah, sure enough I am using a non-released version. Thanks, will fix.

shacker: Fixed.

shacker:
@heymonkeyriot Even though not all pages currently display body, it can be turned it on or off just by adding or removing body in `content_panels` per-model. The payoff is greater code consistency, the ability to deal with your pages in similar ways, and a good demo of composability. I'd be willing to remove body if you guys don't like it, but this felt like a positive step to me. Maybe I should demo the ability by having one of the index pages have some body text, and show the added content panel there?

heymonkeyriot:
@shacker are we able to remove body from the content_panels if we're doing `content_panels = BasePageFieldsMixin.content_panels + [{model specific fields}]`? If we can that's pretty cool and not something I've come across before. I'd very much like to not have a body section on any index page, I think it goes against the principals of Create Once Publish Everywhere, but if there's a way to have the `BasePageFieldsMixin` with body, and then not have it visible on the IndexPage models I'm absolutely fine with this. I think we may need @daaray again to make a casting vote :)

daaray:
Yes, you can mutate the content_panels.  We have an example of this in [BlogPage](https://github.com/wagtail/bakerydemo/blob/master/bakerydemo/blog/models.py#L64). As long as the `body` field is not a required field, this should be relatively trivial.

shacker:
It's actually simpler than the interesting mutation we have on `BlogPage` - just specify the content panels you want on the model normally. So fixing this was as simple as using this on the index pages:

```
    content_panels = Page.content_panels + [
        FieldPanel('introduction', classname="full"),
        ImageChooserPanel('image'),
    ]
```

Interesting aside though - I first tried doing a fancy mutation (similar to the one on BlogPage) on BreadsIndex to remove the  body content panel. It worked, but had the side effect of also hiding the body content panel on other pages as well. No idea why and definitely not expected. But it's clearer and more explicit to just ask for the panels you need, so no matter.

PR #93, opened 2017-03-06
Tweak choices definitions for Operating Hours

### LINKED ISSUE #86: OperatingHours defines weekday abbreviations twice

daaray:
Fixes #86


bakerydemo/locations/models.py
```diff
@@ -20,20 +20,20 @@ class OperatingHours(models.Model):
     """
     MONDAY = 'Mon'
```
shacker: Not sure I follow - weren't we going to delete this duplicate set (and tweak whatever code needed to support the de-duplication?)

shacker:
De-duped in added commit.

daaray:
This changes things a little bit.  I am not necessarily opposed, but here was my initial thinking with my revised implementation:

- Constants are defined once, and utilized via reference in other portions of the app
- The entry in the DB should be a shortened version of the day
- The entry in the dropdown, when editing, should be the full version of the day

This revision removes the notion of a defined CONSTANT for each day. 

As an example, if this app were built out more, and we had a query that filtered by day, I would rather see:

`OperatingHours.objects.filter(day=OperatingHours.MON)`

vs

`OperatingHours.objects.filter(day='MON')`

shacker:
A shortened version of the day is stored in the DB, and it can still be utilized from elsewhere in the app with `from foo.models import DAY_CHOICES`.  Oh - I'll move the choices outside the model to make that easier. But yes, then the query syntax would be `OperatingHours.objects.filter(day='MON')` . Better than duplicating the set of choices - imagine if there were 200 on the list - you don't want the possibility of them going out of sync.

heymonkeyriot:
Not sure it's helpful, but I'm agnostic on this either way so happy with whatever you guys decide on :)

shacker:
I'm not agnostic on this one at all - the current implementation is super-non-dry and non-standard - when I first saw it I thought it was an oversight, not a feature. Hence the reason for this ticket to begin with.

shacker:
I'm going to be a bit of a jerk and merge this myself, then move the choices out in a separate commit, so they're reusable (which is a legit concern).

PR #104, opened 2017-03-15
Aria roles and labels across templates

heymonkeyriot:
I've probably added fewer roles and labels than I should do, but I think this is at least an improvement and does mean that it's possible to keyboard tab navigate the site and improve the accessibility of it.


shacker:
This is great @heymonkeyriot  - thanks!

PR #112, opened 2017-03-23
Remove body from content_panels on LocationPage

### LINKED ISSUE #111: Remove page body field from content_panels on Locations
We're getting by fine with short introduction and a map  and not using body in the demo. Remove?

shacker:
Resolves #111

### DIFF
--- a/bakerydemo/locations/models.py
+++ b/bakerydemo/locations/models.py
@@ -121,7 +121,10 @@ class LocationPage(BasePageFieldsMixin, Page):
     ]
 
     # Editor panels configuration
-    content_panels = BasePageFieldsMixin.content_panels + [
+    content_panels = [
+        FieldPanel('title', classname="full"),
+        FieldPanel('introduction', classname="full"),
+        ImageChooserPanel('image'),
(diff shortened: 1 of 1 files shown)


heymonkeyriot:
I've started adding in body copy to these pages as they feel quite stubby. There's a couple of them that also feel like the introductory on some of them are slightly too long. I think we should retain the body text here.

shacker:
I personally feel the location pages don't need much info - short and sweet - just the location, m'aam. But I'll defer to you on this one. We can kill the PR.

heymonkeyriot:
Oh, you've merged the PR. It's a little odd with the design as there's a lot of things in a relatively small amount of space. I personally think we should reinstate.

shacker:
I merged it with a 2nd commit which reverted the first change and added target = _blank.

shacker:
Oh, sorry, I was referring to #113 above. Yeah this merge looks like a mistake, sorry. Will reinstate the field tonight.

shacker:
Body field on LocationPage is restored in latest PR.

PR #113, opened 2017-03-23
Remove map links from Locations index

shacker:
- Maps are already in location detail views - these lead readers out rather than in


heymonkeyriot:
I'd be in favour of keeping these here but instead adding a `_blank`. I think that's better from a UX point of view where we're allowing users to look at various map locations from the list view, whilst avoiding it being an entirely outward journey.

shacker:
OK

PR #115, opened 2017-03-23
Clean up blockquote handling

shacker:


bakerydemo/templates/blocks/blockquote.html
```diff
@@ -1,7 +1,10 @@
 {% load wagtailimages_tags %}
 
-
-<blockquote><p class="text">{{ self.text }}</p>
-<footer><p class="attribute_name">{{ self.attribute_name}}</p></footer>
+<blockquote>
+    <p class="text">
+        {{ self.text }}<br />
+        {% if self.attribute_name %}
```
heymonkeyriot: This would need to be in a footer to be semantically correct

bakerydemo/templates/blocks/blockquote.html
```diff
@@ -1,7 +1,10 @@
...
+            &mdash; {{ self.attribute_name}}
```
heymonkeyriot: This is purely for a styling effect right? I think if it is we should be doing that via a CSS before statement

bakerydemo/templates/blocks/blockquote.html
shacker: Hmm, I thought it was unusual to use `footer` in a blockquote, but you should go ahead and do whatever you like on this one.

heymonkeyriot:
@shacker would you be happy to close this in favour of #123?

PR #117, opened 2017-03-23
Styling amendments from Holland

heymonkeyriot:
Ready for review.

A number of minor amendments based on feedback and comments from folks here at the wagtail.space sprint.


shacker:
This is great, thanks Edd.

PR #119, opened 2017-03-26
Remove BasePageFieldsMixin

shacker:
- Return to explicit fields everywhere

While remove this and returning to explicit fields everywhere, I grappled a bit with the fact that we had/have two levels of abstraction: 1) `BasePageFieldsMixin`, and 2) `base.StandardPage`. 

`StandardPage` was also inheriting from `BasePageFieldsMixin` but once I made those fields explicit on StandardPage, I realized that we could have most of the models inherit from StandardPage instead. I'm not sure that would be any cleaner - we would again be more DRY, but it might lead back to the potential confusion for new developers. Probably best to leave it as shown in this PR, but just noting it for feedback here.


heymonkeyriot:
This looks good to me. Thanks for taking it on Scot! I think it makes sense to be explicit despite the additional verbosity that I was worried about. Perhaps something that we can note in docs/ inline comments may be that you can abstract these fields back to a mixin?

I've added a very small commit to organise the panels so that they're intro, image and streamfield. Since the streamfield is quite long on a couple of them they were getting slightly lost underneath.

@tomdyson and @daaray it'd be great to have your input on this too.

daaray:
👍  from me.  I see this as another opportunity for a follow-on blog post explaining how one could utilize mixins for similar cases.

tomdyson:
+1 from me too. I think you've made the right trade-offs here. Thanks, @shacker!

PR #121, opened 2017-03-27
CSS refactoring

### LINKED ISSUE #81: CSS refactor
The CSS isn't in great shape because I've been adding new styles to the end of the file for each new feature branch. We should refactor before it to make it easier to navigate and maintain.

heymonkeyriot:
Closes #81 

This PR

 - Removes unnecessary vendor prefixes (e.g. on linear-gradient and border radius)
 - Organises common styles under relevant headers, with an index at head of file
 - Removes styles that are now un-used in the project
 - Alphabetises all properties

It doesn't try to reduce duplication in any meaningful way (there were one or two places where I removed it) and instead sticks with the verbosity to make it easy to follow. Something to do in the future perhaps.


heymonkeyriot:
(Slightly cheating by adding these deletions to this branch, but seemed easier to test on a single branch rather than two. These are residual JS and CSS files from the original base theme that were refactored out of existence during the sprint)

shacker:
Looks like a merge conflict to resolve on this one (and github is not showing most of the diffs, prob due to the merge conflict).

heymonkeyriot:
Ah, it's because the blockquote PR got merged first. I should have put a note on that one. Will resolve the conflicts on this one though.

shacker:
Tower of Power of CSS @heymonkeyriot  - thanks!

PR #123, opened 2017-03-28
Blockquote amend

heymonkeyriot:
Ironically the `::before` element causing us the problem was from Bootstrap so I owe myself an apology for believing it was my fault. I've gone for the most simple way of differentiating between the quote and attribution possible.

image

### DIFF
--- a/bakerydemo/base/blocks.py
+++ b/bakerydemo/base/blocks.py
@@ -42,7 +42,7 @@ class BlockQuote(StructBlock):
     """
     text = TextBlock()
     attribute_name = CharBlock(
-        blank=True, required=False, label='e.g. Guy Picciotto')
+        blank=True, required=False, label='e.g. Mary Berry')
 
     class Meta:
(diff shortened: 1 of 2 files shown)


bakerydemo/base/blocks.py
```diff
@@ -42,7 +42,7 @@ class BlockQuote(StructBlock):
     """
     text = TextBlock()
     attribute_name = CharBlock(
-        blank=True, required=False, label='e.g. Guy Picciotto')
+        blank=True, required=False, label='e.g. Mary Berry')
```
heymonkeyriot: As much as I love Fugazi I'm not sure Guy Picciotto is appropriate as an attribute name example. Have changed to Mary Berry who's a celebrity baker here in the UK

shacker: Big Mary Berry fan here! (the great british baking show is a hit here as well)

PR #127, opened 2017-03-29
List regression on homepage

heymonkeyriot:
I think this relates to #107 that I clearly should have tested more thoroughly. Apologies.


heymonkeyriot:
Should be good to go now @shacker

PR #136, opened 2017-04-02
Have images on initial Heroku deployment

heymonkeyriot:
Currently we don't map the image location to Heroku's media system. I confess that how images are served by Heroku remains a mystery to me, given that I thought they only allowed a through relationship to an S3 bucket for serving static content. In any case this appears to work until the app goes to sleep, at which point images are no longer displayed as expected (which I believe is the expected behaviour of Heroku so would make sense)

@daaray I think you're the only other one with some experience of Heroku. Would you be able to have a review of this if you have the chance?


shacker:
I'll leave this PR to @daaray's input, but the fact that we're recommending a host that loses your images unless you set up media on a totally separate host seems frankly nuts to me. Sorry to be blunt, but this seems like  nothing but additional hassle for demo experimenters.

daaray:
Heroku provides, arguably, one of the easiest deployment options for people that need to show this off to others in a non-local environment.  That does not take away from the fact that it is impossible to work around Heroku's ephemeral file system without a 3rd party solution, such as S3 integration.

I am still of the opinion that offering a well-documented workflow for getting this demo up and running at a publicly available URL is important, as many demo-ers will need to share to present this to other stakeholders.  From that perspective, Heroku still seems to be the lowest hanging fruit.

heymonkeyriot:
I agree with @daaray on this too. At the moment Heroku offers by far and away the easiest way to get an initial sense of the project, and something that someone of almost any technical background could deploy. The ephemeral file system is definitely irritating since when the app goes to sleep the images get lost but that seems okay.

@daaray are you okay that this solution does at least populate those images on first deployment. I've realised I've been really foolish deleting the fork I'd made to test this. I can't see a way to test this without making a fork and amending, making this code change and amending the deployment link within the readme.

daaray:
@heymonkeyriot, I am definitely OK with that.  Anything that further facilitates getting the user a demoable site at a public URL relatively pain-free is a boon.

PR #138, opened 2017-04-02
Add favicon and remove unused images from static img folder

heymonkeyriot:
image

A bread favicon for our bakery. I've also removed some images from the very beginning of the project that I think we'd forgotten about.


shacker:
Nice, thanks. When I purged the unused images the first time I did it from an `Image` query and didn't consider that we had images in the dir that weren't referenced in the admin. Thanks!

PR #141, opened 2017-04-02
Amend readme to improve flow of the document

heymonkeyriot:
I've tried to structure the readme so it's slightly easier to follow. It's now split between

 - installation
 - next steps
 - contributing
 - notes on the demo

I've tried to be neutral in the different installation options. Beyond the initial statement that for newbies Vagrant will be easier I don't think there's a good reason for us to weigh-in one way or the other. I've also slightly downplayed the "we don't recommend using this for your site..." style of comments, they're important to have in there, but as the second sentence it seemed slightly off-putting. I think we should also be striving for the code quality in the demo being good enough that we wouldn't be embarrassed if we saw it on a different site.

I haven't, but I wonder whether it's worth including a screenshot (or screenshots) within the readme?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,34 +1,39 @@
 Wagtail demo project
 =======================
 
-This is a demonstration project for [Wagtail CMS](http://wagtail.io).
+This is a demonstration project for the amazing [Wagtail CMS](https://github.com/wagtail/wagtail).
 
-*We do __not__ recommend using this project to start your own site*. This project is only to provide some examples of
-implementing common features, it is not an exemplar of Django or Wagtail best practice.
+The demo site is designed to give some example of common features and recipes to introduce you to Wagtail. Beyond the code it will also let you explore the admin and editorial interface of the CMS.
 
-This project can be installed in one of three ways:
+Note we do _not_ recommend using this project to start your own site but intend it to be a springboard for you to get started.
 
-- Vagrant
-- Docker
-- Traditional/manual Django setup
+**Document contents**
 
-If you're new to Python/Django, we suggest you run this project on a Virtual Machine using Vagrant or
-via Docker, both of which help resolve common software dependency issues. Developers more familiar with
-virtualenv and traditional Django app setup instructions should use the Local Setup instructions below.
+- [Installation ](#installation)
+- [Next steps](#next-steps)
+- [Contributing](#contributing)
+- [Other notes on the demo](#other-notes-on-the-demo)
 
-Once you're familiar with the examples in this project and you want to start a real site, run
-the ``wagtail start`` command in a fresh virtual environment, explained in the
-[Wagtail CMS Documentation](http://wagtail.readthedocs.org/en/latest/getting_started/).
+# Installation
+
+- [Vagrant](#setup-with-vagrant)
+- [Docker](#setup-with-docker)
+- [Virtualenv](#setup-with-virtualenv)
+- [Heroku](#deploy-to-heroku)
+
+If you're new to Python and/or Django, we suggest you run this project on a Virtual Machine using Vagrant. If you're more comfortable 
+with Docker then you can setup with Docker. Both Vagrant and Docker will help resolve common software dependency issues. Developers more familiar with
+virtualenv and traditional Django app setup instructions can use the instructions within the [Setup with virtualenv](#setup-with-virtualenv) section below.
 
 Setup with Vagrant
 ------------------
 
-### Dependencies
-* [Virtualenv](https://virtualenv.pypa.io/en/stable/installation/)
-* [VirtualenvWrapper](http://virtualenvwrapper.readthedocs.io/en/latest/install.html) (Optional)
+#### Dependencies
+* [Vagrant](https://www.vagrantup.com/)
+* [Virtualbox](https://www.virtualbox.org/)
 
-### Installation
-Run the following commands:
+#### Installation
+Once you've installed the necessary dependencies run the following commands:
 
 ```bash
 git clone git@github.com:wagtail/bakerydemo.git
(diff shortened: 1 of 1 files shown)


shacker:
I made some additional edits and tweaks, and will leave some additional notes in the text now.

readme.md
```diff
@@ -1,34 +1,38 @@
...
-This project can be installed in one of three ways:
+Note we do _not_ recommend using this project to start your own site - the demo is intended to be a springboard for you to get started. Feel free to copy code from the demo into your own project.
 
-- Vagrant
-- Docker
-- Traditional/manual Django setup
+**Document contents**
 
-If you're new to Python/Django, we suggest you run this project on a Virtual Machine using Vagrant or
-via Docker, both of which help resolve common software dependency issues. Developers more familiar with
```
shacker: I'm curious how Docker/Vagrant resolve software dependency issues differently or better than the standard and Django-recommended approach of using pip in a virtualenv? What makes them better or easier than the recommended approach?

readme.md
```diff
@@ -1,34 +1,38 @@
...
-the ``wagtail start`` command in a fresh virtual environment, explained in the
-[Wagtail CMS Documentation](http://wagtail.readthedocs.org/en/latest/getting_started/).
+# Installation
+
+- [Vagrant](#setup-with-vagrant)
+- [Docker](#setup-with-docker)
+- [Virtualenv](#setup-with-virtualenv)
+- [Heroku](#deploy-to-heroku)
+
+If you're new to Python and/or Django, we suggest you run this project on a Virtual Machine using Vagrant or Docker (whichever you're most comfortable with). Both Vagrant and Docker will help resolve common software dependency issues. Developers more familiar with
```
shacker: Same question - how does this suggestion that a user go off and learn about another technology stack, then install and configure it before trying our demo, help the user?

readme.md
```diff
@@ -105,23 +113,16 @@ To set up your database and load initial data, run the following commands:
...
+If you don't want to test locally you can deploy a demo site to a publicly accessible server with [Heroku's](https://heroku.com)
+one-click deployment solution to their free 'Hobby' tier:
 
 [![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy?template=https://github.com/wagtail/bakerydemo)
 
 If you do not have a Heroku account, clicking the above button will walk you through the steps
-to generate one.  After which, you will be presented with a screen to configure your app. For our purposes,
+to generate one.  At this point you will be presented with a screen to configure your app. For our purposes,
 we will accept all of the defaults and click `Deploy`.  The status of the deployment will dynamically
 update in the browser. Once finished, click `View` to see the public site.
```
shacker: I'm confused by the Heroku recommendation. Why do we choose them rather than any of the other  Django hosts out there? Especially since we go on to warn that your images will probably disappear the next day? How is this acceptable?  And now they have to go set up S3 to solve that problem? Seems like we're recommending a big hassle. I also still don't see how hosting is a concern for our demo site - we're recommending that users NOT start their projects from our codebase - which means they'll only be running our demo locally. Deployment docs all over the web, without favoring a single host. The Django docs have their own links. I know I'm in the minority on this, but it just feels so far away from our layer of concern, favors one host over others, introduces additional dependences (S3)...

readme.md
```diff
@@ -158,37 +159,47 @@ environment, set the same environment variables there using the method appropria
...
-### Instructions for demo developers preparing this archive for distribution
+Please optimize any included images to 1200px wide with JPEG compression at 60%. Note that `media/images` is ignored in the repo by `.gitignore` but `media/original_images` is not. Wagtail's local image "renditions" are excluded in the fixture recipe above.
 
-If you change content or images in this repo and need to prepare it for export, do the following on a branch:
+Make a pull request to https://github.com/wagtail/bakerydemo
 
-```
-./manage.py dbshell
-delete from django_session;
-delete from wagtailimages_rendition;
```
shacker: I removed these lines because the included fixtures recipe already excludes the sessions and renditions table, so this was redundant.

readme.md
heymonkeyriot: Vagrant is definitely an easier way to bring up a build than virtualenv. To use virtualenv one requires a reasonably good knowledge of how Python is running on your MacOSX and can cause software conflicts/problems if done incorrectly (because it's using global dependencies). Specific to this project if a user doesn't have a non-system version of Python (e.g. via Pyenv or Brew), PIP, Postgres etc all installed (or if something's gone wrong with their `$PATH` variable) they'll find it frustrating bringing up a build.

Since Vagrant (and to an extent Docker) specify the whole machine it's much more self-contained and entirely agnostic about the host computer, which should make it easier for a new user to implement it. It should be a case of `vagrant up` and waiting a while. Speaking for myself, even though each Virtualbox VM can be several GB, conceptually I still much prefer Vagrant. It I destroy a vagrant box I know it's gone. In any case given that we're offering four ways to bring up a build I'm not sure we're prejudicing people if they want to use virtualenv to bring up the demo, though happy to tweak the language if you think we should?

readme.md
heymonkeyriot: To bring up an initial build on Heroku the user doesn't need to have S3 set-up, it's only if they wanted it to persist that they'd need to do that. I have to say it's a pretty seamless deployment (I was surprised dealing with it yesterday for the image issue just how straightforward it is) and is certainly something I could imagine suggesting a client do if they were interested in exploring Wagtail. I don't think we can presume that people will always want, or be able, to deploy locally and Heroku is currently our best option. Hopefully we can offer others in the future though.

On the note around starting projects from our codebase, I think Heroku would make this more difficult. The deployment is tied to our git repo, so at the very least, if the user would have to fork the repo before they started using it for their own thing, which hopefully would encourage them to not do that and start from scratch based on their own needs.

shacker: Fair enough. I guess it's a trade-off between, on the one hand:

- Easy deploy on Heroku but with hassle of setting up separate media hosting on another host

vs.

- Slightly more complicated deploy on WebFactor or Digital Ocean or any of a bunch of other hosts, with persistent media hosting built in.

Anyway, this is a nice readme enhancement. Merging, thanks (assuming there are no issues with the changes I introduced).

PR #143, opened 2017-04-15
Fix media files in Docker image

azza-bazoo:
At the moment, when starting bakerydemo with the Docker setup instructions, media files don't work and the homepage doesn't have any images.

I think there are two problems, both fixed here:
1. images aren't served from uWSGI (I've followed the change from 97409097552d6d31ec0e25bfce790b9a94a02f92)
1. images are owned by root (due to Docker's COPY command), but the code needs write access and runs as uid 1000

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,4 +41,4 @@ RUN DATABASE_URL=postgres://none REDIS_URL=none /venv/bin/python manage.py colle
 
 # start uWSGI, using a wrapper script to allow us to easily add more commands to container startup:
 ENTRYPOINT ["/code/docker-entrypoint.sh"]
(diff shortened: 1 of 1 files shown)


klinger:
Can confirm this bug - with latest code you just get an 500 internal server error caused by not readable images when you click on Breads, Locations, Blogs,... only "about" works well. 

With these PR you see the page as it should be.

shacker:
Nice, thank you! @heymonkeyriot or @daaray can you have a look at this? (I'm not a Docker person).

daaray:
Sorry for the delay in reviewing this PR. Looks great!

PR #147, opened 2017-05-07
Mobile styling amends

heymonkeyriot:
This PR:

- Gives a global mobile menu
- Hides child items on the global mobile menu
- Ensures the hero image doesn't obscure the navigation bar on mobile
- Uses a consistent hero include on gallery page
- Uses a consistent 12-col grid on search page

image

Technically speaking `a730d7` and `e81f54e` apply beyond the mobile styling but seemed overkill to create a separate PR for those minor items.

### DIFF
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -662,10 +662,12 @@ span.outline {
 /* Bread styles */
 /* Bread listview */
 .bread-list-item {
+  display: block;
   border: 1px solid rgba(0,0,0,0.1);
   border-radius: 3px;
   margin: 12px;
   overflow: hidden;
+  color: inherit;
 }
 
 .bread-list-item h2 {
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -396,6 +396,12 @@ li.has-submenu a.allow-toggle {
     border-top: 1px solid rgba(255,255,255,.1);
     padding: 20px 0;
   }
+  li.has-submenu a.allow-toggle {
+    float: none;
+  }
+  li.has-submenu a.caret-custom {
+    display: none!important;
+  }
 }
 /* Custom page footer */
 .footer {
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -749,6 +749,14 @@ span.outline {
   display: block;
   min-width: 450px;
   max-width: 450px;
+  padding: 5px;
+}
+@media (max-width: 766px) {
+  .form-page input, textarea, select {
+    border: 1px solid #ccc;
+    max-width: 100%;
+    min-width: 100%;
+  }
 }
 
 .form-page li input[type=checkbox], input[type=radio] {
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -322,22 +322,22 @@ nav {
   border-top: 1px solid transparent;
 }
 /* The following is to stop a pixel shift on hover */
-.nav-pills>li.breads {
+.nav-pills>.breads {
   width: 90px;
 }
-.nav-pills>li.locations {
+.nav-pills>.locations {
   width: 140px;
 }
-.nav-pills>li.blog {
+.nav-pills>.blog {
   width: 86px;
 }
-.nav-pills>li.gallery {
+.nav-pills>.gallery {
   width: 115px;
 }
-.nav-pills>li.contactus {
+.nav-pills>.contactus {
   width: 148px;
 }
-.nav-pills>li.about {
+.nav-pills>.about {
   width: 88px;
 }
 .nav-pills>li.active>a, .nav-pills>li.active>a:focus, .nav-pills>li.active>a:hover,
@@ -387,10 +387,10 @@ li.has-submenu a.allow-toggle {
 }
 /* Mobile menu styling */
 @media (max-width: 768px) {
-  .nav-pills>li {
+  .nav-pills>.presentation {
     display: block;
     float: none;
-    width: 100%!important;
+    width: 100%;
   }
   .nav-pills> li > a {
     border-top: 1px solid rgba(255,255,255,.1);
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -751,10 +751,10 @@ span.outline {
   display: block;
   width: 450px;
   padding: 5px;
+  border: 1px solid #ccc;
 }
 @media (max-width: 766px) {
   .form-page input, textarea, select {
-    border: 1px solid #ccc;
     width: 100%;
   }
 }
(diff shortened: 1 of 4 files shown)


shacker:
This is great @heymonkeyriot , thanks. Just noticed one minor issue when viewing through Chrome's mobile emulator - the fields on the Contact page aren't visible until you click into them. Otherwise, good to go.

heymonkeyriot:
Ah sorry @shacker I never spotted your comment. Will merge in master here, have a look at fixing that issue and then update.

heymonkeyriot:
Curiously I think this might be a Chrome inspector tool thing. In any case I've added the a 1px border to input fields on the form page for any viewport below 766px (the mobile breakpoint we were using through the project).
image

heymonkeyriot:
@thibaudcolas it looked like you started looking at this and then stopped. Any chance you'd be able to have a quick look at the changes above? Would be cool to get merged in as the demo really doesn't work very well on mobile currently in a few different places.

thibaudcolas:
@heymonkeyriot yep, that's exactly what happened 😁. I'll review it again shortly.

bakerydemo/static/css/main.css
```diff
@@ -702,7 +725,7 @@ span.outline {
 }
 
 .bread-detail figure img {
-  width: auto;
+  height: auto;
```
thibaudcolas: This makes the image enormous on the square-ish breakpoints I usually use. Should probably be defined only for a specific breakpoint.

Before/after:

mobile-amends-image

bakerydemo/static/css/main.css
```diff
@@ -665,7 +688,7 @@ span.outline {
   padding: 0;
 }
 
-.bread-list-item .image:hover img {
+.bread-list-item:hover img {
```
thibaudcolas: This should be defined on hover on the `a` tags rather than the whole div – otherwise it looks as if the hovered area is clickable even though it's not. Or otherwise make this `bread-list-item` element the `a` tag.

See how the pointer cursor is only on specific parts of the list item:

mobile-amends-hover

bakerydemo/static/css/main.css
```diff
@@ -427,6 +445,11 @@ li.has-submenu a.allow-toggle {
 .breadcrumb-container + content > .hero {
   margin-top: -36px;
 }
+@media (max-width: 768px) {
+  .breadcrumb-container + content > .hero {
+    margin-top: 0px;
+  }
```
thibaudcolas: Some issues:

- There is no `content` tag in HTML.
- From what I can see this is fixing the `margin-top: -36px` breaking the layout on mobile.
- (0 should be unitless)

This would be easier to achieve by changing the `.breadcrumb-container` definition above to be `position: absolute;` and `left: 0; right: 0;`, to produce the same effect without having to change the styles of multiple elements. Then the two `.breadcrumb-container + content > .hero` can be removed.

bakerydemo/static/css/main.css
```diff
@@ -385,6 +385,24 @@ li.has-submenu a.allow-toggle {
 .caret-custom:after {
     content: "▼"!important;
 }
+/* Mobile menu styling */
+@media (max-width: 768px) {
+  .nav-pills>li {
+    display: block;
+    float: none;
+    width: 100%!important;
```
thibaudcolas: The `!important` shouldn't be necessary. Just change the `.nav-pills>li.breads` (and others) width definitions to only use the class name, not element, and then change this declaration to target `.nav-pills > .presentation`.

bakerydemo/static/css/main.css
```diff
@@ -385,6 +385,24 @@ li.has-submenu a.allow-toggle {
 .caret-custom:after {
     content: "▼"!important;
 }
+/* Mobile menu styling */
+@media (max-width: 768px) {
+  .nav-pills>li {
+    display: block;
```
thibaudcolas: As far as I can see Bootstrap already does this in `navs.less`.

bakerydemo/static/css/main.css
```diff
@@ -385,6 +385,24 @@ li.has-submenu a.allow-toggle {
...
+    float: none;
+    width: 100%!important;
+  }
+  .nav-pills> li > a {
+    border-top: 1px solid rgba(255,255,255,.1);
```
thibaudcolas: As far as I can see this only changes the color, the border is already defined elsewhere – so this should only override `border-top-color`.

bakerydemo/static/css/main.css
```diff
@@ -385,6 +385,24 @@ li.has-submenu a.allow-toggle {
...
+    padding: 20px 0;
+  }
+  li.has-submenu a.allow-toggle {
+    float: none;
+  }
+  li.has-submenu a.caret-custom {
+    display: none!important;
+  }
```
thibaudcolas: I don't see an example of those `.has-submenu` rules happening in the demo, could you point me to how to get this?

bakerydemo/static/css/main.css
```diff
@@ -338,7 +338,7 @@ nav {
   width: 148px;
 }
 .nav-pills>li.about {
-  width: 98px;
+  width: 88px;
```
thibaudcolas: 👍

bakerydemo/static/css/main.css
heymonkeyriot: They're not in the demo fixtures but if you were to add some of the sub-pages to the menu (via the 'Show in menus' in 'promote') and they'll be visible. This caret was an incredibly hacky way of presenting the dropdown on desktop. (The menu's constructed by base/templatetags/navigation_tags.py)

heymonkeyriot:
Thanks for reviewing @thibaudcolas !

I'd forgotten just how awful this CSS file is to edit. There was a decision made to not include any front-end tooling within the demo at the sprint, which as you say makes life a little tricky. Will run through `allthethings` either with comments or amended code 🙌

bakerydemo/static/css/main.css
heymonkeyriot: I'm not even sure why I've amended a hover on mobile styling `:facepalm:`

Will go and look at the hover though because you're right it makes it confusing.

thibaudcolas: TBH I don't think it brings _that much_ to the UI to make it orange-ish like that (it's images of bread...), so this could simply be removed.

bakerydemo/static/css/main.css
thibaudcolas: Gotcha 👍. Just reviewed that change, works as expected.

thibaudcolas:
@heymonkeyriot I fixed some of these while reviewing, you can cherry-pick the commits from https://github.com/thibaudcolas/bakerydemo/commits/139-mobile-styling if you want. NB: that branch is based off the `wagtail-2.0` one so I could review, but that shouldn't affect those commits.

thibaudcolas:
Looking at this again, the PR fixes more issues than it introduces, and it looks unlikely to be picked up again. So – merging.

PR #158, opened 2017-10-04
Add instructions & requirements for populating S3 bucket with demo data

tobiasmcnulty:
It turns out `DJANGO_SETTINGS_MODULE` must be set in the Heroku/Dokku environment (rather than just in the production WSGI file), because it's also needed for running management commands (and `manage.py` presumably should continue to point to the dev rather than production settings).

There is one minor improvement that we could make instead, which is to provide a mechanism for populating the S3 bucket with the demo media.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -159,6 +159,11 @@ environment, set the same environment variables there using the method appropria
 
 Once Heroku restarts your application or your Docker container is refreshed, you should have persistent media storage!
 
+To copy the initial data included with this demo to the S3 bucket (assuming you ran `./manage.py load_initial_data` per
+the above), you can use the AWS CLI included with the requirements:
+
+    heroku run aws s3 sync bakerydemo/media/original_images/ s3://<bucket-name>/original_images/
+
 # Next steps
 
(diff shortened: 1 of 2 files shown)


daaray:
The [Heroku django template](https://github.com/heroku/heroku-django-template/blob/master/manage.py) indicates it should be set in manage.py:

https://github.com/wagtail/bakerydemo/blob/master/manage.py#L11

Unsure how to best reconcile off the top of my head.

tobiasmcnulty:
I think `manage.py` may not come into play in this case because uWSGI is loading the production wsgi file directly: https://github.com/wagtail/bakerydemo/blob/master/Procfile#L2

Also, `manage.py` is loading the `dev` rather than `production` settings module

tobiasmcnulty:
@daaray Ohh, it's coming from `app.json`: https://github.com/wagtail/bakerydemo/blob/master/app.json#L8

Which is apparently ignored by Dokku: http://dokku.viewdocs.io/dokku/advanced-usage/deployment-tasks/#example-appjson

Perhaps I should remove it from `app.json` along with this PR to avoid the duplication? I can't think of a use case for someone wanting to change that environment variable in Heroku at the time the app is created...can you?

PR #172, opened 2017-11-06
Allow 'vagrant' username to be configured by passing DEV_USER variable

gasman:
This allows the script to work on official Ubuntu images, which use the username 'ubuntu' instead.


heymonkeyriot:
Looks good to me.

PR #177, opened 2017-12-01
Make dj shortcuts work from any directory

balazs-endresz:
Based on the `wagtail-2.0` branch, which was the default one within `vagrant-wagtail-develop`.


gasman:
-1. Let's not do this, please... having a global `DJANGO_SETTINGS_MODULE` hiding in the environment really messes you up when you create additional projects within the same VM, which is something I often do when trying to create minimal test cases for bugs (and something we actively encourage when reporting bugs on Github).

balazs-endresz:
Sure, I didn't know that.

PR #178, opened 2017-12-01
Disable password validators locally

balazs-endresz:
Based on the `wagtail-2.0` branch, which was the default one within `vagrant-wagtail-develop`.

### DIFF
--- a/bakerydemo/settings/dev.py
+++ b/bakerydemo/settings/dev.py
@@ -6,3 +6,5 @@ EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 # BASE_URL required for notification emails
 BASE_URL = 'http://localhost:8000'
+
+AUTH_PASSWORD_VALIDATORS = []


gasman:
Not sure if this is something we want to encourage - for a demo project like this it's probably harmless enough, but in a real-world project it leaves open the possibility of insecure accounts being created during development and being forgotten about when the project is initially deployed to staging/production.

balazs-endresz:
Isn't that true for `DEBUG = True` and using the email backend as well?
Not a big issue of course, it's just that we have to type a bit more when creating users and logging in locally.

gasman:
I'd say that `DEBUG = True` and the email backend are a different situation. To put those into production, you have to actively make a mistake in configuration - i.e. specifying dev settings instead of production. In the case of insecure passwords, it's possible for them to end up on the production server just through laziness or forgetfulness:

* user starts a development site and creates a user account with an insecure password (as permitted by the dev settings)
* user creates some content locally
* user decides to put the site into production, using the content they've entered locally, and uploads a database dump of their development site
* production site now includes an insecure account, despite having validation to disallow them

balazs-endresz:
Yeah, I see what you mean now. I didn't think of that scenario.

PR #182, opened 2018-01-12
Set site name to Wagtail Bakery

gasman:
'Main' isn't really suitable as a site name, because it results in a limited-permission user (e.g. an editor with edit permission on the Blog section only) receiving the welcome message "Welcome to the Main Wagtail CMS".


gasman:
Merged in f5ee3939578b4b2042f70c538098220b873dd32f

PR #184, opened 2018-01-18
Cleanup Search Form Attributes

rgant:
Remote empty id attribute from search input. Remove _lpchecked attribute from search form because it isn't part of standard HTML. https://stackoverflow.com/questions/8879303/what-is-lpchecked-1-for-in-a-form


gasman:
Thanks again! Merged in fd37ac93101bcb53c8cc31c46e5ab3d683f9f621.

PR #188, opened 2018-02-20
Upgrade vm.box_version to 2 and beyond

ethankent:


gasman:
Thanks @ethankent - well spotted!

PR #191, opened 2018-02-26
Add missing Docker Compose dependency to readme file

blgo:
Docker Compose does not come installed with Docker. This change might save a few minutes of confusion and Google to some users willing to try the demo.


gasman:
Thanks @blgo!

PR #192, opened 2018-02-26
Fix broken link to 'Getting started' page in sample about page

blgo:


blgo:
'Getting started' Link was using a version specific URL for http://docs.wagtail.io/ which was not working.
I replaced the version by 'latest' which will always work.

gasman:
Thanks @blgo!

I'd suggest linking to [stable](http://docs.wagtail.io/en/stable/getting_started/index.html) rather than latest - the 'latest' version corresponds to git master, which may contain changes that aren't in a release yet.

blgo:
Thanks @gasman for the quick response.
I agree using stable is better than latest, however, I decided to use latest because in most URLs in this project is written "docs.wagtail.io/en/latest".
I have push the change with stable to this PR.

PR #204, opened 2018-05-25
Docker: add zlib-dev

jorrit-wehelp:
Got an error while building through docker-compose.
Pillow was compiling (didn't get a precompiled wheel...?),
and it was complaining about missing zlib headers, so added it to the libraries list in Dockerfile.


paolodina:
this breaks continuous integration as well.

kaedroho:
Thanks!

PR #216, opened 2018-07-26
Docker will now force-remove rendition entries from db

Eeemil:
Since Docker will use a new file system on a rebuild, old renditions in
$MEDIA/images will no longer be present. By removing rendition entries from db
Wagtail will be forced to generate new renditions, instead of providing dead
links to the old renditions.

### DIFF
--- a/bakerydemo/base/management/commands/delete_renditions.py
+++ b/bakerydemo/base/management/commands/delete_renditions.py
@@ -1,0 +1,14 @@
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+# Force removal of wagtail image rendition-entries in db will force wagtail to
+# regenerate renditions. Useful if renditions are removed, for instance when
+# using Docker where the file system is renewed on each build.
+
+class Command(BaseCommand):
+    def handle(self, **options):
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT count(*) FROM wagtailimages_rendition")
+            row=cursor.fetchone()
+            cursor.execute("DELETE FROM wagtailimages_rendition")
+            print("{} renditions removed".format(row[0]))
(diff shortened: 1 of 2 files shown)


cnk:
@Eeemil Sorry this PR did not get reviewed sooner. The management command to clear renditions from the database is really helpful. 

I am not 100% sure about running it from docker-entrypoint.sh. It looks to me like the Dockerfile is trying to be an example of how one might build production images (starting with the slim image and clearing up caches as you go). If I interpreted this correctly, then I am concerned about code that always clears the image renditions when the container starts (even though it is just fine for the demo). 

Can you help me think through when this command should run - and when it will run? When I stopped my containers using Ctrl-C to stop `docker-compose up`, the database and the image renditions were both preserved. `docker-compose up -d` followed by `docker-compose stop` also preserved the database and rendition files. Using `docker-compose down` got rid of the database (and presumably the files). 

I think the need for your delete_renditions command comes when the database data is preserved by the app container is not. I think that combination does not occur when using the instructions and Dockerfile + docker-compose.yml in this project. It happens when using the Dockerfile and docker-compose.yml in https://github.com/wagtail/docker-wagtail-develop (which does preserve the database volume). In which case we need to add the management command here and then use it from https://github.com/wagtail/docker-wagtail-develop.

Amondale:
@Eeemil you're right:
**The management command to clear renditions from the database is really helpful.**

My thinking is that Dockerizing any file-based app or set of apps is a "shifting sands" proposition, and rather than welding in a command to clear what may or may not be a bunch of orphan renditions, why not just ship it from github and document it along with the use cases that make it a valuable utility. "load_initial_data" and "delete_renditions" are needed in the early dev ("hack") phases of experimenting with the bakerydemo/Wagtail learning experience.

PR #222, opened 2018-10-27
Bump awscli to be compatible with Py37

nott:
`awscli` version we used to pin depends on `pyyaml` that doesn't support py37

The issue got fixed in https://github.com/aws/aws-cli/pull/3430

`botocore` is also updated as `awscli` and `botocore` should usually be synchronized (before doing that I was getting HTTP 500 on Heroku).


gasman:
Thanks @nott!

Issue #223, opened 2018-10-28
Is elasticsearch 2.4 supported by AWS t2.micro.elasticsearch?


shurph:
In AWS docs the maximum allowable version for `t2.micro.elasticsearch` is **2.3**:
https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-supported-instance-types.html

But currently bakery demo is using version **2.4**: https://github.com/wagtail/bakerydemo/blob/563d580/requirements/base.txt#L3

Unfortunately I don't have access for now to any AWS console to check this, but it seems like a potential issue.

gasman:
2.4.1 is the Python library version, and as I understand it the API is stable across major releases, so using the 2.4.1 library with a 2.3 server should be OK.

In any case, I suggest that we don't bother supporting EC2 for bakerydemo at all. As far as I can see, it's not been documented beyond a few notes in settings/production.py; we aren't actively testing it when we make changes to the code; and now that we have one-click setup for both Heroku and Divio Cloud, I don't think there's any reason why anyone would choose to try out the demo through EC2. The notes might be useful for people deploying their own sites on EC2, but bakerydemo isn't meant to be a guide to deploying on every possible platform. (Of course, if people want to write their own deployment guide using bakerydemo as an example, that's fine...)

I'd even go further and say that we could drop the instructions for hosting static files on S3, or rather move them to somewhere outside bakerydemo itself. If people need their media files to persist over multiple sessions, that probably means they're doing something beyond trying it out as a demo.

shurph:
@gasman,
>  If people need their media files to persist over multiple sessions, that probably means they're doing something beyond trying it out as a demo.
Or they may use it to test this functionality during some development.

But at the same time I totally agree with you that worrying about supporting EC2 for bakery demo is not the first priority thing (even it's not the second priority :-) ).

gasman:
Closing as there's no action to be taken here - the EC2 elasticsearch config in settings/production.py isn't being actively tested, but it's not doing any harm to keep it in there.

PR #225, opened 2018-10-28
Move elasticsearch requirements to an appropriate place

shurph:
- elasticsearch is used only on production settings by default
- for development purposes elasticsearch is installed by a `provision.sh` script in a separate repo: https://github.com/wagtail/vagrant-wagtail-develop


shurph:
BTW, I think we need to decide which version we should use in bakery demo: 2.3 or 2.4. Details are there: https://github.com/wagtail/bakerydemo/issues/223

PR #226, opened 2018-11-08
Correct the html element main

TravellerSam:
Correct the HTML5 element 'content' to 'main'. 'content' is obsolete and wasn't meant to be used in ordinary HTML. 
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/content


gasman:
Thanks @TravellerSam!

PR #232, opened 2019-03-20
Enable CSP

kaedroho:
This PR adds a ``Content-Security-Policy`` header to responses from the Wagtail admin. This should prevent any unsafe eval statements from running so we can catch these issues early.


thibaudcolas:
Closing in favour of #423. There seem to be quite a few conflicts so I suspect we will need to restart this from scratch anyway.

PR #233, opened 2019-06-29
Update Python to 3.7 in Dockerfile

johnraz:
Upgrade of the python version, depends on #235

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.5-alpine
+FROM python:3.7-alpine
 
 ADD requirements/ /requirements/
 RUN set -ex \


johnraz:
Hey @thibaudcolas, thanks for the feedback.

I don't have much experience with Heroku myself but it seems to be supported according to their changelog: https://devcenter.heroku.com/changelog-items/1442

Should I proceed and update `runtime.txt` ?

`expat-dev` was required to build the C bindings of lxml IIRC.

Thanks!

gasman:
Judging from #234 the need to add `expat-dev` is a separate fix, unrelated to the Python upgrade.

johnraz:
ah cool @gasman thanks for clarifying, it was indeed confusing me that upgrading Python would require another dependency... Will split this up in 2 PR's then 👍

PR #241, opened 2019-07-27
Keep the year set to the current year

DanielSwain:
Let's keep the bread as fresh as possible.


KalobTaulien:
Hey @DanielSwain 

This looks good to me. Thanks for the contribution and keeping the bread fresh!

Issue #247, opened 2019-08-10
Unclear about Wagtail Features in the demo app


supersexy:
It is not clear if the [demo app](https://github.com/wagtail/bakerydemo) is meant to show **all** features of Wagtail in a "kitchen sink demo" style or if it is missing important features. 

The demo "About" page itself contains a list of features that is very helpful, but also at the end the two words "lots more", which is absolutely not helpful.

It would be very important for new developers to better understand the feature set that comes with Wagtail to have **precise** documentation about the features that are used in the demo.

Also I would like to suggest to make the demo a "full featured kitchen sink" demo to have a place that shows all the Wagtail features available out of the box. 

Also it would be great to have a "kitchen sink extended" demo that showcases all the Wagtail features plus the "official" extensions that can be seen in some repos of the wagtail github organization. I can see these repos but it is not easy to understand the status of these repos - why do they exist, why are they not part of Wagtail? Or are they? Unclear.

thibaudcolas:
Hey @supersexy 👋 

This demo site is not meant to demonstrate all of the features of Wagtail, it’s only meant to demonstrate common features for a relatively small site. I would assume "lots more" in the About page refers to things that are not worth listing because they are relatively minor.

We’ve discussed _how much_ the bakerydemo should cover in the past and decided it shouldn’t be a "kitchen sink", as this would go counter to providing a meaningful introduction to newcomers who want to see how a real site is built (https://github.com/wagtail/bakerydemo/issues/175#issuecomment-350566374).

Personally I think it would be really cool to have more "kitchen sink" demos, but bakerydemo isn’t meant for that. For other repositories in https://github.com/wagtail, none of them are "official" extensions, they are either extensions that are worked on by a large number of people involved with the development of Wagtail itself, or work-in-progress for things that might get into Wagtail. If you find the status of any specific project unclear I would suggest making an issue on that project.

rblcoder:
Hi, @thibaudcolas are the changes up to the mark?

thibaudcolas:
Thanks @rblcoder 🙂.

PR #249, opened 2019-09-07
Move demo app features from about page to readme

rblcoder:
#247


thibaudcolas:
Thank you @rblcoder! That will help a lot of people I’m sure.

PR #254, opened 2019-10-04
include procps in RUN_DEPS

tobiasmcnulty:
It's tiny, and pretty handy for debugging.


tobiasmcnulty:
Added half a meg to the image size, for the record:

Screen Shot 2019-10-04 at 3 56 09 PM

PR #255, opened 2019-10-06
copy files from fixtures dir to MEDIA_ROOT (or remote storage) during load_initial_data

tobiasmcnulty:
If we're not using a local storage, we can copy images to the remote storage ourselves to make it easier for the user.

Tested on Google Cloud, AWS, and locally.


gasman:
It would probably be a good thing for this copy to happen on local storage too (perhaps keeping the initial files in `base/fixtures`) - currently bakerydemo suffers from the quirk that `media/original_images` can't be gitignored.

(Making that change does have the downside that anyone doing a git pull on an existing instance will have those image files deleted - if that's a concern we could add a startup check that copies them back...)

tobiasmcnulty:
@gasman Good idea, I've updated this.

tobiasmcnulty:
Hey @gasman, did you want to review this again before merging?

PR #261, opened 2019-10-22
Make site work offline by self-hosting webfonts, jQuery, Bootstrap etc.

### LINKED ISSUE #242: Demo site should work without an internet connection
Without an internet connection, the bakerydemo website looks broken. This is because it relies on CDNs to load its JS (jQuery, Bootstrap), stylesheet (Bootstrap), fonts (Google Fonts), and icons (Font Awesome).

I think it would be important for a basic demo site not to assume users will have a (stable) internet connection, and work standalone.

Here is what it looks like at the moment:

bakerydemo-offline

---

The solution is most likely to vendor all of the above dependencies into the project. I think for the fonts it might be ok not to vendor (since there are fallback fonts) – but at the same time might as well have everything loading as expected offline.

con-cat:
Fixes #242 

I've brought the following in as static files:
- Bootstrap
- jQuery
- HTML5 Shiv
- Respond.js
- Google fonts: Alegreya, Lato, Lora and Open Sans (font and css files generated via the very handy [google-webfonts-helper](https://github.com/majodev/google-webfonts-helper))

For the Font Awesome Icons, I've switched to using the wagtailfontawesome package that's already included in the project's requirements to render them in the front end.

Here's a [screenshot](https://user-images.githubusercontent.com/10053862/67257381-69b31200-f4d7-11e9-8c25-234677def647.jpg) of the site working with no internet connection in an incognito window after these changes.


con-cat:
Hi @tobiasmcnulty, I've added those extra static font files and the build is passing now :)

tobiasmcnulty:
@con-cat Thanks! This looks great. One last thing: I'm having trouble verifying `respond.min.js`; could you point me to where you downloaded it and/or download a fresh copy from a CDN or GitHub? The version in this PR seems slightly different than what I'm seeing for v1.4.2 elsewhere. If you enable 'allow edits from maintainers' I'm happy to make this change and merge, too (see: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

Otherwise, I've confirmed the site works offline for me too and looks identical (as far as I can tell) to the site using the CDN-hosted assets.

con-cat:
I believe I downloaded it from Github here: https://github.com/scottjehl/Respond/blob/master/dest/respond.min.js

I've got 'allow edits from maintainers' enabled if it isn't the right file. Thanks!

tobiasmcnulty:
Sure enough - thanks! For whatever reason it looks like there have been some minor tweaks to that since the original 1.4.2 release. 🤷‍♂ 

Thanks for the contribution!

PR #266, opened 2020-05-10
height :auto, for displaying images without distortion.

jonndoe:
When resizing browser window, images get some distortion, when seen from mobile its doesnt look nice. Suggest to auto height,  for displaying images of stream field in blog page without distortions.


gasman:
Thanks @jonndoe!

PR #268, opened 2020-06-04
Allow all hosts in dev mode to simplify cross-browser testing

thibaudcolas:
This is needed to test with the demo site on mobile devices / Browserstack. In the past I’ve added this in my `local.py`, but it’s annoying to have to re-add this when resetting my environment, and I assume it doesn’t hurt to have this directly in the project’s settings.


marteinn:
I see no danger in this, I always do the same in my local environments, as long as they dont slip up to production.

thibaudcolas:
Thank you both!

PR #275, opened 2020-08-16
upgrade to Wagtail 2.10

cnk:
Upgrade bakerydemo to use Wagtail 2.10 and Django 3.1 

I elected to use wagtail.contrib.legacy.richtext so as not to have to change '.rich-text img' style in main.css. Would people like that to be done differently?


gasman:
Thanks @cnk!
I think it would be good to eliminate the .rich-text style, since people often borrow code from bakerydemo and I wouldn't really want wagtail.contrib.legacy.richtext to be common practice. As far as I can recall, most or all of the rich text areas where we'd want to include images are StreamField rich text blocks, so that should give us an alternative container element to attach the style to - that can be done in a separate PR though.

We should also update our initial-data fixture to populate the new `slug` field on form fields, so that we aren't relying on the system check to fill it in on first startup (which will fail when we drop the unidecode library in 2.12).

PR #278, opened 2020-10-05
Add Wagtail SEO to Experiments Branch

NathHorrigan:
This PR adds the [Wagtail SEO Library](https://docs.coderedcorp.com/wagtail-seo/) for internal testing and evaluation.

## Changes Made:
- Migration added that replaces any pages with a `owners` field of None with the root user.
- Wagtail SEO added as Submodule.
- SEO Options added to HomePage, Standard Page, Breads Index Page and Breads Page.
- Large Twiter card style added to Bread Page.
- Article SEO type added to Bread Page and Standard Page.
- Wagtail SEO Markup added to `base.html`


Vagrantfile
```diff
@@ -74,6 +74,27 @@ Vagrant.configure(2) do |config|
   # Enable agent forwarding over SSH connections.
   config.ssh.forward_agent = true
 
+  # Workaround to prevent missing linux headers making new installs fail.
```
kaedroho, line 77: Probably be better to just update the base image on master so this wouldn't be necessary?

gasman:
I'm assuming that the motivation for this was Torchbox-specific and the need for it has now passed, so it no longer needs to be merged - let me know if not.

Issue #284, opened 2020-12-04
Building the docker image fails with current pip


WhyNotHugo:
```console
$ docker-compose up --build
[...]
+ python3.7 -m venv /venv
+ /venv/bin/pip install -U pip
Collecting pip
  Downloading pip-20.3.1-py2.py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.1.1
    Uninstalling pip-20.1.1:
      Successfully uninstalled pip-20.1.1
Successfully installed pip-20.3.1
+ /venv/bin/pip install --no-cache-dir -r /requirements/production.txt
Collecting aws-requests-auth==0.4.0
  Downloading aws-requests-auth-0.4.0.tar.gz (9.0 kB)
Collecting boto3==1.9.189
  Downloading boto3-1.9.189-py2.py3-none-any.whl (128 kB)
INFO: pip is looking at multiple versions of aws-requests-auth to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r /requirements/production.txt (line 10) and botocore==1.12.33 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested botocore==1.12.33
    boto3 1.9.189 depends on botocore<1.13.0 and >=1.12.189

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
ERROR: Service 'app' failed to build : The command '/bin/sh -c set -ex     && BUILD_DEPS="         build-essential         git         libexpat1-dev         libjpeg62-turbo-dev         libpcre3-dev         libpq-dev         zlib1g-dev     "     && apt-get update && apt-get install -y --no-install-recommends $BUILD_DEPS     && python3.7 -m venv /venv     && /venv/bin/pip install -U pip     && /venv/bin/pip install --no-cache-dir -r /requirements/production.txt     && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false $BUILD_DEPS     && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 1
```

This is due to the new pip [dependency resolver](https://pyfound.blogspot.com/2020/11/pip-20-3-new-resolver.html).

[...]

PR #285, opened 2020-12-22
fix requirements mismatch

JNaftali:
When trying to follow the instructions for docker, there was an error while installing dependencies - boto3 required a higher version of botocore than we were installing. I made the smallest change I could manage (allowing more recent versions of botocore 1.12.x) and it fixed the problem.

### DIFF
--- a/requirements/production.txt
+++ b/requirements/production.txt
@@ -11,7 +11,7 @@ boto3==1.9.189
 google-cloud-storage==1.20.0
 django-storages>=1.8,<1.9
 # For retrieving credentials and signing requests to Elasticsearch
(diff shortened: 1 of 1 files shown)


joakimnordling:
Thanks! This was what I needed to be able to try it out.

JNaftali:
Just looked back at this and realized that (a) this is a potential solve for #284 (I did an Open Source Sin and didn't check for an open issue before I opened this PR) and (b) the travis CI build failed but I do not understand the error. I apologize on both counts

gasman:
Thanks @JNaftali!

PR #286, opened 2021-02-02
Update runtime.txt

tahirpoduska:
Update supported runtime for heroku: https://devcenter.heroku.com/articles/python-support

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.7.6
+python-3.8.7


thibaudcolas:
Thank you for looking into this @tahirpoduska. I decided to merge #288 instead, as I don’t think we would want to switch to Python 3.8 just yet. And if we did, we’d also need to update the project’s Dockerfile at least.

Issue #287, opened 2021-02-04
Heroku python runtime not found


BrianXu20:
This appears when I try to deploy it on Heroku (Manually & one-click deployment):

![Screen Shot 2021-02-04 at 1 43 28 PM](https://user-images.githubusercontent.com/73645085/106852906-cd287980-66f3-11eb-9d77-7ede76af9edf.png)

And this is the official Supported runtimes document:
https://devcenter.heroku.com/articles/python-support#supported-runtimes

![Screen Shot 2021-02-04 at 2 15 49 PM](https://user-images.githubusercontent.com/73645085/106853076-14166f00-66f4-11eb-9165-de3e2576bc74.png)

The runtime specified in the runtime.txt is no longer supported by Heroku.

I changed the python runtime from 3.7.6 to 3.7.9 then it worked

thibaudcolas:
Fixed in #288.

PR #288, opened 2021-02-04
Heroku python runtime not available issue 

BrianXu20:
Issue: https://github.com/wagtail/bakerydemo/issues/287
Change the python runtime from 3.7.6 to the Heroku supported 3.7.9 version

### DIFF
--- a/runtime.txt
+++ b/runtime.txt
@@ -1,1 +1,1 @@
-python-3.7.6
+python-3.7.9


thibaudcolas (approved):
Thank you @BrianXu20! I had a look around at our other Python version definitions and the Dockerfile is the only one that’s up-to-date, with `python:3.7` so matches this change.

PR #292, opened 2021-04-23
Add Gitpod

allcaps:
Test this PR:
1. Go to https://github.com/allcaps/bakerydemo/#setup-with-gitpod
2. Click the Gitpod button
3. Relax

Note: The url in the readme points to my fork. This url needs to be updated on merge. `/allcaps/` => `/wagtail/`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -46,7 +46,7 @@ Setup a development environment and run this demo website with a single click (r
 
 [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/allcaps/bakerydemo/)
 
-Once Gitpod has started, click the preview url to open the website in a new tab.
+Once Gitpod has fully started and a preview of the bakery website has appeared in the "Simple Browser" panel, click the arrow button to the right of the URL bar to open the website in a new tab.
 Go to ``/admin/`` and login with ``admin / changeme``.
 
 Setup with Vagrant
--- a/readme.md
+++ b/readme.md
@@ -31,13 +31,16 @@ This demo is aimed primarily at developers wanting to learn more about the inter
 
 # Installation
 
+- [Gitpod](#setup-with-gitpod)
 - [Vagrant](#setup-with-vagrant)
 - [Docker](#setup-with-docker)
 - [Virtualenv](#setup-with-virtualenv)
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -39,6 +39,16 @@ This demo is aimed primarily at developers wanting to learn more about the inter
...
 virtualenv and traditional Django app setup instructions should skip to [Setup with virtualenv](#setup-with-virtualenv).
 
+Setup with Gitpod
+-----------------
+
+Setup a development environment and run this demo website with a single click (requires a Github account):
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/allcaps/bakerydemo/)
+
+Once Gitpod has started, click the preview url to open the website in a new tab.
```
gasman, line 49: Suggesting a bit more detail in the instructions, from the perspective of someone who didn't entirely know what to expect from Gitpod startup...
```suggestion
Once Gitpod has fully started and a preview of the bakery website has appeared in the "Simple Browser" panel, click the arrow button to the right of the URL bar to open the website in a new tab.
```

gasman (changes requested):
This works really nicely for me!

We might consider updating some of the surrounding text:

* For the intro "If you're new to Python and/or Django...", I think we could recommend Gitpod as the first option - something like "If you want to see what Wagtail is all about, we suggest trying it out through Gitpod. If you want to set up Wagtail locally instead, and you're new to Python and/or Django..."
* The Heroku instructions probably shouldn't start with "If you don't want to test locally", since we've already given people a non-local option. In fact, is there any real reason to keep Heroku around as an alternative at this point?

allcaps:
Thank you for the review!

I applied the requested changes.

> * The Heroku instructions probably shouldn't start with "If you don't want to test locally", since we've already given people a non-local option. In fact, is there any real reason to keep Heroku around as an alternative at this point?

Agreed, the `not locally` is confusing. Heroku is publicly available and persistent, not focused on code. We should keep it. I changed the wording.

gasman (approved):

gasman:
Thanks @allcaps! Merged in 4469a5a182f3c34db520979bd257a9a5cc4620fa.

PR #293, opened 2021-05-18
Use Wagtail 2.13 for bakery demo and GitPod.

KalobTaulien:


requirements/base.txt
```diff
@@ -1,4 +1,4 @@
 Django>=3.1,<3.2
 django-dotenv==1.4.1
-wagtail>=2.12,<2.13
+wagtail>2.12,<=2.13
```
gasman, line 3: Ah, just noticed this isn't quite right... `>2.12,<=2.13` will allow 2.12.1 but disallow 2.13.1 (if/when that exists).

PR #294, opened 2021-07-16
Upgrade to Django 3.2

gasman:
Set DEFAULT_AUTO_FIELD to prevent deprecation warnings (Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField')


lb-:
Looks good to me, merging in.

PR #295, opened 2021-07-20
Update urlconf to use django-2-style paths

gasman:


zerolab (approved):
❤️ little changes like this

PR #304, opened 2021-09-09
Update all Django docs references to the latest (3.2) version

lb-:
* many of the Django docs links were pointing to v1, updated all to v3.2


gasman (approved):
Thanks @lb-!

PR #309, opened 2021-12-26
Remove null=True from CharField

### LINKED ISSUE #308: Title in HomePage model should be not null
## Issue
Avoid using `null` on string-based fields such as CharField and TextField. 
If a string-based field has null=True, that means it has two possible values for `no data`: NULL, and the empty string.
* See: https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null

This is also followed in Wagtail, check this fix: https://github.com/wagtail/wagtail/commit/eb9cff7bf388735e988bfd9f084d7a5d34f0ba42

In this demo, I notice the following four fields violate this practice:
`/bakerydemo/base/models.py`
```python
class HomePage(Page):
    promo_title = models.CharField( null=True, )
  featured_section_1_title = models.CharField(null=True, )
  featured_section_2_title = models.CharField(null=True, )
  featured_section_3_title = models.CharField(null=True, )
```

## Fix
* Remove `null=True`. 
* Do relevant changes, like make the default be empty string.

vihaanthora:
Closes #308 

- [x] Removed null=True from models.CharField present in the file bakerydemo/base/models.py
- [x] Migrated the change


jacobtoppm (changes requested):
Thanks @vihaanthora! Would you be able to give the migration a more explanatory name, please? Otherwise looks good to me.

vihaanthora:
Thank you for reviewing the pull request @jacobtoppm! I have made the requested change. Please let me know if anything else needs to be changed.

gasman:
Thanks @vihaanthora!

PR #310, opened 2022-01-07
Add a util script to create large amounts of random data

RealOrangeOne:
It's very useful to have not just a small wagtail site to work with, but potentially large scales of data. This script can automatically create data, to allow for more extreme load / performance testing, both of components in the admin, main site utils and the main request/response cycle.

The foundations are here, interested in some more feedback on what else could be added.

This depends on #311 to work as expected.


tomdyson:
Nice one @RealOrangeOne! 

See also https://github.com/tomdyson/wagtail-fakenews, which takes a slightly different approach with streamfield and image generation.

gasman (approved):

PR #311, opened 2022-01-07
Fix initial page tree

RealOrangeOne:
This was previously creating a single `Page`, with multiple "specific" pages pointing to it.

### DIFF
--- a/bakerydemo/base/fixtures/bakerydemo.json
+++ b/bakerydemo/base/fixtures/bakerydemo.json
@@ -3383,7 +3383,7 @@
   }
 },
 {
-  "model": "blog.blogindexpage",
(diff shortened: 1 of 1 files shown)


gasman (changes requested):
Good catch! I don't think the fix here is right, though - there's [an existing breadsindexpage entry with id=3](https://github.com/wagtail/bakerydemo/blob/e94bfbbc994bb3fa6eb2c58dbaec931acd15cedb/bakerydemo/base/fixtures/bakerydemo.json#L3246-L3252), so this change causes that one to be overwritten and the introduction field lost. I think this entry should just be deleted - does that sound ok to you?

gasman (approved):

PR #320, opened 2022-02-08
Fix release phase

fabienheureux:
[Django 4.0 removed `django-admin.py` entry point. ](https://docs.djangoproject.com/en/4.0/releases/4.0/#features-removed-in-4-0)
This causes bakery demo heroku's postdeploy script to fail. 

This fixes that


gasman:
Thanks @fabienheureux!

PR #326, opened 2022-03-26
Remove elasticsearch

vossisboss:


gasman:
Thanks @vossisboss! I think there are a few other references to elasticsearch we can remove now too:

* [requirements/production.txt](https://github.com/wagtail/bakerydemo/blob/master/requirements/production.txt) - looks like some of the boto / AWS stuff could possibly go too, but it's probably a deeper task to untangle all of that...
* [search/views.py](https://github.com/wagtail/bakerydemo/blob/4fb21cd9f8e1c5528fc9a6e45f510330c5bea6fc/bakerydemo/search/views.py#L17-L36) - currently there are two different code paths depending on whether Elasticsearch is in use or not, but the simple `Page.objects.live().search` version should work in all cases now. There's also [a note about this in the readme](https://github.com/wagtail/bakerydemo#note-on-demo-search), which can be dropped too if that does indeed work.
* [settings/production.py](https://github.com/wagtail/bakerydemo/blob/4fb21cd9f8e1c5528fc9a6e45f510330c5bea6fc/bakerydemo/settings/production.py#L48-L86) - a bunch of Elasticsearch config that's no longer needed (and also [a brief reference to it in settings/base.py](https://github.com/wagtail/bakerydemo/blob/4fb21cd9f8e1c5528fc9a6e45f510330c5bea6fc/bakerydemo/settings/base.py#L176) that can also go)

gasman:
All working nicely, so I'll go ahead and merge - the other bits of cleanup can be done as a follow-up.

PR #329, opened 2022-04-07
Make the people in bakerydemo more diverse

KalobTaulien:
Out of the 4 original `People` snippets, 100% of them were male. 

This PR adds diversity to the default people when running `load_initial_data`. 

I also changed the images and compressed them using TinyPNG as to not bloat the repo.


kaedroho:
Are these images your own work? If not, could you provide some information on their license?

Otherwise, looks good to me, thanks @KalobTaulien!

KalobTaulien:
@kaedroho Nope, those are royalty free images I found, with no attribution required. :)

PR #330, opened 2022-04-11
Add basic simple_translation setup to bakerydemo

thibaudcolas:
This only adds the required settings, not any database-level changes. I think this is helpful as-is nonetheless, to take some of the guesswork out of setting up translations for testing.


zerolab (approved):
LGTM

zerolab:
@thibaudcolas for a translateable snippet you could have something like:

```python
@register_snippet
class TranslateableSnippet(TranslatableMixin, models.Model):
    text = models.CharField(max_length=255)

    class Meta:
        unique_together = ("translation_key", "locale")

    def __str__(self):
        return self.text


class MyPage(Page):
    # ...
        translateable_snippet = models.ForeignKey(
        "blog.TranslateableSnippet",
        null=True, blank=True,
        on_delete=models.SET_NULL,
        related_name='+'
    )

    content_panels = Page.content_panels + [
       # ...
        SnippetChooserPanel("translateable_snippet"),
    ]
```

bakerydemo/settings/base.py
```diff
@@ -183,3 +185,12 @@
 
 # Wagtail settings
 WAGTAIL_SITE_NAME = "bakerydemo"
+
+USE_I18N = True
```
zerolab, line 189: this is already defined on https://github.com/wagtail/bakerydemo/blob/44cfef774d4680a3b5c5e389da6b3ea81d016640/bakerydemo/settings/base.py#L150

kaedroho:
If we're going to make snippets translatable, we should bootstrap them so that existing installations of bakerydemo can be migrated.

See: https://docs.wagtail.org/en/stable/advanced_topics/i18n.html#making-snippets-with-existing-data-translatable

PR #331, opened 2022-04-25
Updated uses of StreamField to use JSONField

PaarthAgarwal:
I have edited the stream fields of Bakery demo to work on Wagtail 3.0 as suggested by @cnk in https://wagtailcms.slack.com/archives/C81FGJR2S/p1650812809244809?thread_ts=1650802381.876709&cid=C81FGJR2S.

### DIFF
--- a/bakerydemo/base/models.py
+++ b/bakerydemo/base/models.py
@@ -122,7 +122,7 @@ class StandardPage(Page):
         help_text='Landscape mode only; horizontal width between 1000px and 3000px.'
     )
     body = StreamField(
-        BaseStreamBlock(), verbose_name="Page body", blank=True
+        BaseStreamBlock(), verbose_name="Page body", blank=True, use_json_field=True
     )
     content_panels = Page.content_panels + [
         FieldPanel('introduction', classname="full"),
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @PaarthAgarwal! I've now merged this as a new `wagtail-3.0` branch. The `use_json_field` change in Wagtail creates a problem for bakerydemo, since at the current moment it's not possible to write a StreamField definition that works on the latest released version (2.16) and current main (4.0a0, which drops the backwards compatibility for missing `use_json_field`) - for that reason, I'll also revert the change on main so that the backwards compatibility remains until Wagtail 5.0.

PaarthAgarwal:
> Thanks @PaarthAgarwal! I've now merged this as a new `wagtail-3.0` branch. The `use_json_field` change in Wagtail creates a problem for bakerydemo, since at the current moment it's not possible to write a StreamField definition that works on the latest released version (2.16) and current main (4.0a0, which drops the backwards compatibility for missing `use_json_field`) - for that reason, I'll also revert the change on main so that the backwards compatibility remains until Wagtail 5.0.

Awesome!

PaarthAgarwal:
@gasman, should the readme be updated to point the new users to the wagtail-3.0 branch?

gasman:
No, because once I make the change described above, the bakerydemo master branch will work with Wagtail main again (just with a deprecation warning), so ordinary users won't need to deal with the new branch.

PaarthAgarwal:
Got it👍🏼

PR #333, opened 2022-05-23
Restyle CSS Across Site

jhancock532:
Updates the CSS of the site to a modern design. Currently in progress, pending design review fixes and review from Thibaud.


bakerydemo/static/css/font-marcellus.css
```diff
@@ -0,0 +1,9 @@
+@font-face {
+  font-family: 'Marcellus';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: local('Marcellus-Regular'), local('Marcellus Regular'),
+       url('../fonts/Marcellus-Regular.ttf')
```
thibaudcolas, line 7: Could we use woff2 only for all fonts?

bakerydemo/templates/base/home_page.html
```diff
@@ -4,121 +4,99 @@
...
+            </div>
+
+            <div class="col-sm-6 promo">
+                {% if page.promo_image or page.promo_title or page.promo_text %}
+                <div class="col-lg-10 promo-text">
+                    {% if page.promo_title %}
+                    <h2>{{ page.promo_title }}</h2>
+                    {% endif %}
+                    {% if page.promo_text %}
+                    {{ page.promo_text|safe }}
```
thibaudcolas, line 51: Could we try removing the safe?

bakerydemo/templates/base/include/footer_text.html
```diff
@@ -1,5 +1,5 @@
 {% load wagtailcore_tags %}
 
-<div class="copyright text-center text-muted" role="note">
+<div class="copyright" role="note">
```
thibaudcolas, line 3: Remove role

bakerydemo/templates/base/include/header-blog.html
```diff
@@ -0,0 +1,23 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+{% if page.image %}
```
thibaudcolas, line 3: So no image no header?

bakerydemo/templates/base/include/header-hero.html
```diff
@@ -3,16 +3,8 @@
...
     {% image page.image fill-1920x600 as image %}
     <div class="container-fluid hero" style="background-image:url('{{ image.url }}')">
-    <div class="hero-gradient-mask"></div>
-        <div class="container">
-            <div class="row">
-                <div class="col-md-7">
-                    <h1>{{ page.title }}</h1>
-                    {% if page.subtitle %}
-                        <p class="stand-first">{{ page.subtitle }}</p>
-                    {% endif %}
```
thibaudcolas, line 13: Should we not retain the subtitle? Is it displayed elsewhere ?

bakerydemo/templates/blog/blog_index_page.html
```diff
@@ -2,48 +2,34 @@
...
-                                {% if blog.date_published %}
-                                    {{ blog.date_published }} by 
-                                {% endif %}
-                                {% for author in blog.authors %}
-                                    {{ author }}{% if not forloop.last %}, {% endif %}
-                                {% endfor %}
-                            </div>
-                        </a>
-                    </li>
+                    {% include "includes/card/blog-listing-card.html" %}
```
thibaudcolas, line 32: Bit surprised here. Makes our cards less reusable and more error-prone to mandate them to always be li?

bakerydemo/templates/blog/blog_page.html
```diff
@@ -3,42 +3,35 @@
...
-                            {{ page.date_published }}
-                        </div>
-                    {% endif %}
+<div class="container">
+    <div class="row">
+        <div class="col-md-8">
+            <div class="blog__meta">
+                {% if page.authors %}
+                <div class="blog__avatars">
+                    {% for author in page.authors|slice:":1" %}
```
thibaudcolas, line 15: Not sure I understand the slice. We don’t display all authors?

bakerydemo/templates/breads/bread_page.html
```diff
@@ -2,60 +2,56 @@
...
-
                         {% if page.introduction %}
-                            <p class="introduction">
+                            <p class="bread-detail__introduction">
                                 {{ page.introduction }}
                             </p>
                         {% endif %}
 
-                        {{ page.body }}
+                        <div class="hidden-md-down">
```
thibaudcolas, line 19: There’s no way to do this layout with the same markup for mobile & desktop?

bakerydemo/templates/breads/breads_index_page.html
```diff
@@ -2,47 +2,21 @@
...
-        </div>
+        <ul class="bread-list">
+            {% for bread in breads %}
+                {% include "includes/card/listing-card.html" with page=bread %}
+            {% endfor %}
+        </ul>
     </div>
 
-    {% if breads.paginator.count > 1 %}
+    {% if breads.paginator.count > 12 %}
```
thibaudcolas, line 15: Count of pages or items?

bakerydemo/templates/includes/footer.html
```diff
@@ -2,26 +2,26 @@
 
 <div class="container" role="contentinfo">
```
thibaudcolas, line 3: Switch to footer tag?

bakerydemo/templates/includes/footer.html
```diff
@@ -2,26 +2,26 @@
...
     <div class="row">
-        <div class="col-lg-8 col-lg-offset-2 col-md-10 col-md-offset-1">
-            <ul class="list-inline text-center">
-                <li>
-                    <a href="https://github.com/wagtail/wagtail">
+        <div class="col-sm-12">
+            <ul class="list-inline">
+                <li class="footer__icon">
+                    <a href="https://github.com/wagtail/wagtail" target="_blank" aria-label="View Wagtail's source code on GitHub">
```
thibaudcolas, line 8: Target blank needs rel noreferer

bakerydemo/templates/breads/breads_index_page.html
jhancock532: Yes, this should be pages - todo

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
+  --light-brown: #87744f;
+  --mid-brown: #825600;
+  --dark-brown: #553801;
+  --orange: #c55302;
+  --dark-orange: #833701;
+  --font--primary: "Marcellus", sans-serif;
+  --font--secondary: "Open Sans", sans-serif;
+}
+
+/* Utilities */
```
thibaudcolas, line 34: Utilities should be the very last thing in this file so they can override component styles. Probably doesn’t matter either way for `u-sr-only` but I think it’d be good for us to follow the pattern nonetheless.

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
+
+/* Body */
+
 body {
+  font-family: var(--font--secondary);
+  font-size: 16px;
```
thibaudcolas, line 52: All font sizes should be defined in `rem` so the text can be resized with browser settings.

```suggestion
  font-size: 1rem;
```

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
+  line-height: 24px;
```
thibaudcolas, line 53: Same for line-height, this should always be defined in proportional units or unitless so it changes according to the font’s size.

```suggestion
  line-height: 1.5;
```

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
+
+/* Body */
+
 body {
+  font-family: var(--font--secondary);
```
thibaudcolas, line 51: I don’t totally get the primary/secondary split if the body (which I assume is for most of the text) is secondary. Could we change the variables so they better represent what the fonts are for? `font--headings`, `font--body` perhaps?

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
 
 p {
-  margin: 0 0 30px;
-  color: #444;
+  font-family: var(--font--secondary);
+  font-style: normal;
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 24px;
+  color: var(--black);
```
thibaudcolas, line 108: Do we actually want text to be a full black? I’m surprised this isn’t a very dark shade of grey like on our other projects.

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
+  }
+}
+
+ol,
+ul {
+  margin-bottom: 30px;
+}
+
+.hero h1 {
+  color: white;
```
thibaudcolas, line 150: Be nice if we were consistent with our variables usage even if it’s for things like "white".

bakerydemo/static/css/main.css
```diff
@@ -14,81 +14,182 @@
...
-a, a:focus, a:visited {
-  color: #d4566b;
+/* Links, Lists, Figures */
+
+a,
+a:focus,
+a:visited {
+  color: var(--orange);
   text-decoration: none;
 }
```
thibaudcolas, line 172: I think this needs to be revisited, at least so links inside text have an underline. Right now it’s very hard to tell links apart from normal text.

bakerydemo/static/css/main.css
```diff
@@ -137,31 +210,37 @@ ul {
...
-  position: absolute;
-  width: 75%;
+  color: var(--grey);
+  font-family: var(--font--secondary);
+  font-size: 14px;
+  line-height: 22px;
+  padding-top: 10px;
 }
+
+@media (min-width: 766px) {
```
thibaudcolas, line 231: Should we use 768 for this for consistency with other styles?

bakerydemo/static/css/main.css
```diff
@@ -235,341 +400,466 @@ cite {
...
+  color: var(--dark);
+  font-family: var(--font--secondary);
+  font-size: 18px;
+  line-height: 28px;
+  padding: 10px 0;
 }
-.nav-pills>.about {
-  width: 88px;
+
+@media (min-width: 1150px) {
```
thibaudcolas, line 519: Could we define breakpoints with CSS variables so it’s clear which ones are consistent and which aren’t? and so they have defined names

bakerydemo/static/css/main.css
```diff
@@ -235,341 +400,466 @@ cite {
...
-.navbar-toggle .icon-bar {
-  background-color: #fff;
+.breadcrumb a:hover {
+  color: var(--dark-orange);
 }
 
 /* Pagination navigation */
-nav[role=pagination] {
-  margin-top: 50px;
+nav[role="pagination"] {
```
thibaudcolas, line 768: There’s no pagination role in HTML so this needs switching to a class.

bakerydemo/static/css/main.css
```diff
@@ -804,232 +1172,325 @@ span.outline {
...
 
-.homepage .home-hero {
-  margin-bottom: 0;
-  padding-bottom: 60px;
-  text-align: center;
-}
 .homepage .home-hero .lead {
-  color: #ddd;
-  font-size: 1.6em;
+  font-family: "Marcellus", serif;
```
thibaudcolas, line 1194: We should consistently use our variables for this (here this uses the wrong fallback).

bakerydemo/static/css/main.css
```diff
@@ -1048,18 +1509,349 @@ span.outline {
...
+
+.blog-listing-card:last-child {
+  padding-bottom: 100px;
+}
+
+.blog-listing-card__link {
+  display: flex;
+  flex-direction: column;
+  padding-bottom: 30px;
+  border-bottom: 1px solid rgba(135, 116, 79, 0.25);
```
thibaudcolas, line 1705: Feels like this color needs to be a CSS variable

bakerydemo/templates/includes/header.html
```diff
@@ -2,33 +2,40 @@
...
+            <a href="/" class="navigation__logo">The Wagtail Bakery</a>
 
-                <form action="/search" method="get" class="search" role="search">
-                    <input name="q" type="text" placeholder="Search the site" aria-label="Search the site" autocomplete="off">
-                    <a href="#" class="search-icon">
-                        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 451 451"><path d="M447.05 428l-109.6-109.6c29.4-33.8 47.2-77.9 47.2-126.1C384.65 86.2 298.35 0 192.35 0 86.25 0 .05 86.3.05 192.3s86.3 192.3 192.3 192.3c48.2 0 92.3-17.8 126.1-47.2L428.05 447c2.6 2.6 6.1 4 9.5 4s6.9-1.3 9.5-4c5.2-5.2 5.2-13.8 0-19zM26.95 192.3c0-91.2 74.2-165.3 165.3-165.3 91.2 0 165.3 74.2 165.3 165.3s-74.1 165.4-165.3 165.4c-91.1 0-165.3-74.2-165.3-165.4z"></path></svg>
-                    </a>
-                </form>
+            <button type="button" class="navigation__mobile-toggle navbar-toggle collapsed" data-toggle="collapse"
+                data-target="#main-navigation" aria-label="Mobile menu" aria-expanded="false">
```
thibaudcolas, line 9:
```suggestion
                data-target="#main-navigation" aria-label="Toggle mobile menu" aria-expanded="false">
```

bakerydemo/templates/includes/header.html
```diff
@@ -2,33 +2,40 @@
...
-                        <ul class="nav nav-pills">
-                            {# main_menu is defined in base/templatetags/navigation_tags.py #}
-                            {% get_site_root as site_root %}
-                            {% top_menu parent=site_root calling_page=self %}
-                        </ul>
-                    </nav>
-                {% endblock %}
-            </div>
+            {% block main_navigation %}
+            <nav class="collapse navbar-collapse" id="main-navigation" role="navigation"
```
thibaudcolas, line 17:
```suggestion
            <nav class="collapse navbar-collapse" id="main-navigation"
```

bakerydemo/templates/includes/header.html
```diff
@@ -2,33 +2,40 @@
...
+                aria-label="Primary site navigation">
```
thibaudcolas, line 18: "navigation in the label is redundant this is is a navigation landmark to start with (it’ll be announced as such).

```suggestion
                aria-label="Main">
```

bakerydemo/templates/includes/header.html
```diff
@@ -2,33 +2,40 @@
...
+                <ul class="navigation__items nav-pills">
+                    {# main_menu is defined in base/templatetags/navigation_tags.py #}
+                    {% get_site_root as site_root %}
+                    {% top_menu parent=site_root calling_page=self %}
+                </ul>
+            </nav>
+            {% endblock %}
+
+            <form action="/search" method="get" class="navigation__search" role="search">
+                <label for="search-input" class="u-sr-only">Search this site</label>
```
thibaudcolas, line 28: The label is useless if there is an `aria-label` on the field as per below.

bakerydemo/templates/tags/breadcrumbs.html
```diff
@@ -3,24 +3,25 @@
...
-            {% endif %}
-          </div>
+          <li><a href="/">Home</a> {% include "includes/chevron-icon.html" with class="breadcrumb__chevron-icon" %}</li>
+          {% for ancestor in ancestors|slice:"1:" %}
+          {% if forloop.last %}
+          <li class="active" aria-level="{{ancestors|length}}">
+            {{ ancestor }}
+          </li>
+          {% else %}
+          <li><a href="{% pageurl ancestor %}" aria-level="{% cycle 1 2 3 4 5 %}">{{ ancestor }}</a>
```
thibaudcolas, line 18: aria-level is for headings, it likely does nothing for links so should be removed.

thibaudcolas (changes requested):

bakerydemo/templates/blog/blog_page.html
jhancock532: This was a design decision that Ben Tuckwell suggested, as displaying multiple authors made the section look cluttered. I have removed the slice to show all the information being pulled from Wagtail 👍

bakerydemo/static/css/font-marcellus.css
jhancock532: I've added this in 👍 I've replaced the ttf with a woff for the fallback font, as this is how the other fonts are loaded in.

bakerydemo/templates/base/home_page.html
jhancock532: Sadly, this causes the following glitch: 

image

bakerydemo/templates/base/include/header-blog.html
jhancock532: Fixed this so blogs can show without hero images 👍

bakerydemo/templates/base/include/header-hero.html
jhancock532: I've double checked all the references to this component, and the subtitle is not part of any of their page models. These pages typically render an `introduction` field instead, which is included elsewhere in their page template.

bakerydemo/templates/includes/footer.html
jhancock532: Having switched to the footer tag, I removed the role attribute, as this is no longer necessary. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/contentinfo_role

bakerydemo/templates/includes/footer.html
jhancock532: Added in!

bakerydemo/static/css/main.css
jhancock532: I've refactored the typography across the site to follow this convention 👍

bakerydemo/static/css/main.css
jhancock532: Refactored to match!

bakerydemo/static/css/main.css
jhancock532: I've replaced this and other instances with `--dark` 👍

bakerydemo/static/css/main.css
jhancock532: Yep, I've added this in.

bakerydemo/static/css/main.css
jhancock532: Unfortunately this isn't possible - I did originally try this earlier in development. Some more information about why: https://bholmes.dev/blog/alternative-to-css-variable-media-queries/

bakerydemo/static/css/main.css
jhancock532: Updated to use variable, interestingly this uses the correct fallback - my original definition uses the wrong one as Marcellus has serifs.

bakerydemo/static/css/main.css
jhancock532: Added as a transparent border variable

bakerydemo/templates/tags/breadcrumbs.html
jhancock532: Removed!

bakerydemo/templates/tags/gallery.html
```diff
@@ -1,12 +1,13 @@
...
-<div class="col-sm-6">
-    <figure class="gallery-figure">
-        <img src="{{img_obj.url}}" class="img-responsive" />
-        <figcaption>{{ img.title }}</figcaption>
+<div class="picture-card">
+    <figure class="picture-card__image">
+        {% image img fill-645x480-c100 %}
+        <div class="picture-card__contents">
+            <h3 class="picture-card__title">{{ img.title }}</h3>
+        </div>
```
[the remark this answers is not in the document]
jhancock532: Yes, any [flow content](https://html.spec.whatwg.org/multipage/dom.html#flow-content) is valid within a figure tag. Further examples of fancy figures - https://html.spec.whatwg.org/multipage/grouping-content.html#the-figure-element

bakerydemo/templates/tags/gallery.html
```diff
@@ -1,12 +1,13 @@
...
-{% image img fill-285x200-c100 as img_obj %}
-<div class="col-sm-6">
-    <figure class="gallery-figure">
-        <img src="{{img_obj.url}}" class="img-responsive" />
-        <figcaption>{{ img.title }}</figcaption>
+<div class="picture-card">
+    <figure class="picture-card__image">
+        {% image img fill-645x480-c100 %}
+        <div class="picture-card__contents">
+            <h3 class="picture-card__title">{{ img.title }}</h3>
```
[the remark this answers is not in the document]
jhancock532: Yes, this is just following the designs - I will remove the h3

bakerydemo/templates/includes/header.html
jhancock532: Makes sense!

bakerydemo/templates/includes/header.html
jhancock532: Yep, makes sense

bakerydemo/templates/base/home_page.html
thibaudcolas: Yes, this means we should be using `|richtext` rather than `|safe`.

bakerydemo/templates/base/include/header-hero.html
thibaudcolas: This is because you switched `blog_page`’s header template from `header-hero.html` to `header-blog.html`. So we’d still need to make sure `header-blog.html` renders the subtitle.

bakerydemo/templates/tags/gallery.html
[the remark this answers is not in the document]
thibaudcolas: Cool!

PR #336, opened 2022-05-28
Upgrade to Wagtail 3

allcaps:
You can test drive agains my repo https://gitpod.io/#github.com/allcaps/bakerydemo


zerolab:
Had a cursory look and they LGTM.
I think we may want to branch out before this PR to something like `wagtail-2.x` while we have the 2.15 LTS.

Personally I use bakerydemo to develop and test the various packages I maintain, as well as test core PRs. Makes for a better experience (and more real too) than the package test apps

Finally, I do wonder if it is time to change to using `FieldPanel` pretty much everywhere in the scope of this PR (as per https://docs.wagtail.org/en/latest/releases/3.0.html#id1)

requirements/base.txt
```diff
@@ -1,5 +1,5 @@
 Django>=4.0,<4.1
 django-dotenv==1.4.1
-wagtail>=2.16,<2.17
+wagtail>=2.16,<4
```
gasman, line 3: This should be `wagtail>=3.0,<4` since the new `from wagtail import ...` lines won't work on 2.16.

PR #342, opened 2022-06-20
Rename People model to Person

gasman:


gasman:
Rebased

laymonage (approved):
Looks good to me, tested on both existing and new setups of the bakerydemo and everything seems to work fine. This one has been bugging me as well, so thanks for making the fix!

PR #352, opened 2022-08-19
Add `.git-blame-ignore-revs` and fix CI

laymonage:
This adds a `.git-blame-ignore-revs` file per https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

Also fix issues with flake8-print by upgrading it to 5.0.0 and add some `# noqa` comments on `print()` statements per https://github.com/wagtail/wagtail/pull/8935

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,18 @@
+# git-blame ignored revisions
+# To configure, run
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# Requires Git > 2.23
+# See https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt
+
+# Reformatting PR was primarily done in #347
+
+# Initial black reformatting
+40696f941ca2a2b30938fbd2bcb8e197a46024d5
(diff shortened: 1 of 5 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,18 @@
...
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# Requires Git > 2.23
+# See https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt
+
+# Reformatting PR was primarily done in #347
+
+# Initial black reformatting
+40696f941ca2a2b30938fbd2bcb8e197a46024d5
+# Initial isort reformatting
+6732b719834fc3a7d72dfd78f684f066a1fcf11f
```
laymonage, line 12: Not sure if this is worth adding as the diff is pretty small: https://github.com/wagtail/bakerydemo/commit/6732b719834fc3a7d72dfd78f684f066a1fcf11f

kaedroho (approved):
Looks good to me, thanks!

PR #354, opened 2022-08-25
Character limit for introduction in blog pages

vossisboss:
I'm changing the introduction field for BlogPage to be a RichTextField and to have a max_length so that we can demonstrate and continually test the character count feature.

### DIFF
--- a/bakerydemo/blog/models.py
+++ b/bakerydemo/blog/models.py
@@ -54,7 +54,9 @@ class BlogPage(Page):
     https://docs.wagtail.org/en/stable/topics/pages.html#inline-models
     """
 
-    introduction = RichTextField(help_text="Text to describe the page", blank=True, max_length=1000)
+    introduction = RichTextField(
+        help_text="Text to describe the page", blank=True, max_length=1000
+    )
     image = models.ForeignKey(
         "wagtailimages.Image",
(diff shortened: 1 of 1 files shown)


vossisboss:
@gasman Would appreciate a review on this when you get a chance since I would really like to point out this feature in a video I'm making.

gasman (approved):
Thanks @vossisboss! For consistency I think we should do this with the intro fields on the other page types - will do this and add the required migrations when I merge.

gasman (changes requested):
Ah, hang on - if we're changing this from TextField to RichTextField, we're also going to have to make the corresponding change to the template (to add the `|richtext` filter).

vossisboss:
@gasman Would you like me to help out with that? I have some Monday.com tickets I need to tackle first, but I can make those changes after.

gasman:
So this is actually a massive can of worms... if we change TextField to RichTextField, then it'll also start supplying its own `<p>` tags, which then don't match the `<p class="index-header__introduction index-header__introduction--blog">` tag that we need to output the nice intro text style. So then we'd also need to rework our CSS...

I suggest we apply the max length to the "promo text" field on the homepage instead - that's already a RichTextField so it won't break anything.

PR #355, opened 2022-08-25
Set max_length on homepage promo text

gasman:
Less disruptive alternative to #354... cc @vossisboss


vossisboss:
I'm in favor of opening fewer cans of worms. This looks good to me @gasman. You need anything else for merging it?

PR #357, opened 2022-09-13
Fix help text for EmbedBlock

gasman:
The URL suggested in the help text for EmbedBlock is incorrect - users should specify the Youtube page URL (`/watch`) to have it translated to an embed snippet via oembed. Passing an `/embed/` URL fails, as the oembed endpoint doesn't recognise these.


zerolab (approved):
LGTM

PR #362, opened 2022-10-19
Add styles for `EmbedBlock`

laymonage:
## Before

Unstyled EmbedBlock, the iframe is too small

## After

Styled EmbedBlock, the iframe now looks normal


thibaudcolas (approved):
👍 looking good (tested with YouTube embeds only).

PR #364, opened 2022-11-01
Call rebuild_references_index on initial load

gasman:
Bump Wagtail version to 4.1 (and Django to 4.1) and add a call to rebuild_references_index as part of the load_initial_data management command, so that the "Used X times" reference counts for images and snippets are accurate.


laymonage (approved):
Looks good to me!

PR #365, opened 2022-11-03
Explain Gitpod prebuild, add link to trigger a fresh build

allcaps:


readme.md
```diff
@@ -48,7 +48,10 @@ Setup a development environment and run this demo website with a single click (r
 [![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/wagtail/bakerydemo/)
 
 Once Gitpod has fully started, and a preview of the bakery website has appeared in the "Simple Browser" panel, click the arrow button to the right of the URL bar to open the website in a new tab.
-Go to `/admin/` and login with `admin / changeme`.
+Append `/admin/` to the URL and login with username `admin` and password `changeme`.
+
+Gitpod uses [prebuilds](https://www.gitpod.io/docs/prebuilds) to reduce wait time, by installing dependencies and running builds **before** you start a new workspace.
+Unfortunately, prebuilds work as a cache. You could get an outdated build. If that is the case, [trigger a new prebuild](https://gitpod.io/#prebuild/https://github.com/wagtail/bakerydemo).
```
Scotchester, line 54: How would one know if they get an outdated build?

zerolab: May be worth adding a nightly GitHub Action that fetches `https://gitpod.io/#prebuild/https://github.com/wagtail/bakerydemo` ? This way we guarantee some freshness

Scotchester: That seems like a good idea, Dan!

allcaps: @zerolab Good idea! Would you be so kind as to submit a PR for that? I'm not too familiar with Github actions / schedules.

Note, we have some other projects that might require this feature too. See: https://gitpod.io/t/wagtail/projects

I'd assume that this current PR is good to merge too. Although the change that people encounter an outdated prebuild after the scheduled action is added is smaller, it is still there for 24h. And mentioning the behaviour is a good thing as nobody seems to be aware ;)

zerolab: 👍 from me on deploying

thibaudcolas: Just trying this now, I got the following error when following the trigger link: "Running prebuilds without a project is no longer supported. Please add your repository as a project in a team."

Going to https://gitpod.io/t/wagtail/bakerydemo/prebuilds and clicking "Run prebuild" seems to work as expected. Do we want to change the instructions?

PR #368, opened 2022-11-17
Add a special page to demo various panels with streamfields

allcaps:
Screenshot 2022-11-17 at 16 31 33


thibaudcolas (commented):
@allcaps could you add more details about what this is for?

If we intend for this to be a testing-only page, I’d really like us to:

- Incorporate more of the work from #231
- And more recently https://github.com/thibaudcolas/bakerydemo/blob/feature/everything-page/bakerydemo/ui_tests/models.py#L61
- Ship this with at least one instance of the page created in fixtures, as populating such a page takes quite a bit of time.
- Add even basic rendering of all of the page’s fields

The fixtures in particular is something I’d really like to see – making sure we put something together that’s a good demonstration of those StreamField features, and stays on the bakery theme, so we can use this kind of page with clients without it looking too bad.

thibaudcolas:
Closing in favour of #380, which incorporates all the scenarios here with the exception of "StreamField inside InlinePanel".

PR #369, opened 2022-11-18
Create better demo environment

RealOrangeOne:
The current Wagtail demo environment is based on a nightly releases of Wagtail, injected into `bakerydemo`. Instead, we should have a more stable, reliable, production-grade deployment of `bakerydemo` which people can use to get a better feel for Wagtail.

Much of this configuration is duplicated from https://github.com/wagtail/guide/pull/47

This is currently deployed to https://demo.wagtail.org/, but I'll setup automated deployment once this is merged.

<details>
<summary>Login</summary>

image

</details>

<details>
<summary>Wagtail Admin</summary>

image

</details>

<details>
<summary>Basic auth error page</summary>

This exists mostly to cover the old demo site.

image

</details>


bakerydemo/base/management/commands/reset_demo.py
```diff
@@ -0,0 +1,30 @@
...
+
+        # 2. Reset database to nothing
+        self.stdout.write("Reset schema")
+        call_command("reset_schema", interactive=False)
+
+        # 3. Rebuild database
+        call_command("migrate", interactive=False)
+
+        # 4. Re-import data
+        call_command("load_initial_data")
```
allcaps, line 30: Should we also call `update_index`? For the search index?

requirements/base.txt
```diff
@@ -2,4 +2,18 @@ Django>=4.1,<4.2
 django-dotenv==1.4.1
 wagtail>=4.1,<5
 wagtailfontawesome>=1.1.3,<1.2
+dj-database-url==1.0.0
 django-debug-toolbar>=3.2,<4
+elasticsearch==5.5.3
+# Additional dependencies for Heroku, AWS, and Google Cloud deployment
```
gasman, line 8: Why have these been moved from production.txt? We don't really want users to have to install these packages if they're just setting up a local build, especially if they have native code dependencies (as psycopg2 does).

bakerydemo/base/management/commands/reset_demo.py
gasman: @allcaps This happens as part of `load_initial_data`, so isn't needed as a separate step here.

I'd suggest that this might also be a good place to change the password of the admin user to something other than `changeme` - or set up a new superuser account - so that people trying the demo don't take this as an invitation to change the password and lock out everyone else.

RealOrangeOne: My intention is to attempt to either prevent the password from being changed, or have a recurring task which forces the password back. I suspect as part of this we'll have a few different users with different permission levels created by default to allow better testing.

requirements/base.txt
RealOrangeOne: I think I was a little trigger-happy when moving these dependencies. I agree a few of them aren't necessary during local development (`psycopg2` being a good example of one which can be a pain on windows).

RealOrangeOne: With that said, whilst vagrant didn't install the production dependencies, the docker install did, as did the virtualenv install path, so even before they weren't very distinct.

bakerydemo/base/management/commands/reset_demo.py
RealOrangeOne: There's now a `reset_admin_password` command which achieves this.

bakerydemo/base/management/commands/reset_admin_password.py
```diff
@@ -0,0 +1,18 @@
+from django.contrib.auth.models import User
```
zerolab, line 1: suggestion: implement https://docs.wagtail.org/en/stable/reference/hooks.html#before-edit-user to prevent the password change in the first place?

RealOrangeOne: I originally did that, but using signals, and it ended up a little brittle. I'll give it another whirl

zerolab: hmm.. I don't think this will work.

nickmoreton:
I know this isn't a 4.0/4.1 upgrade but I've done a few 😄 and spotted some bits and pieces that could be updated.

- [Live Preview Links](https://docs.wagtail.org/en/stable/releases/4.0.html#opening-links-within-the-live-preview-panel) in the base.html
- [FieldPanel Customisations](https://docs.wagtail.org/en/stable/releases/4.0.html#panel-customisations) classname="full" can be removed.
- In [README.md](https://github.com/RealOrangeOne/wagtail-bakerydemo/tree/feature/demo-environment#preparing-this-archive-for-distribution) there are some updates needed for the fixtures command. `wagtailcore.pagerevision` -> `wagtailcore.revision`

gasman:
@nickmoreton I don't think any of those fixes are specific to the demo environment - feel free to open a new issue or PR for those!

nickmoreton:
> @nickmoreton I don't think any of those fixes are specific to the demo environment - feel free to open a new issue or PR for those!

@gasman OK will do 👍

bakerydemo/base/middleware.py
```diff
@@ -0,0 +1,51 @@
...
+from django.conf import settings
+from django.core.exceptions import MiddlewareNotUsed
+from django.template.loader import render_to_string
+
+
+class DemoBannerMiddleware:
+    """
+    When bakerydemo is being run as the Wagtail Demo, show a banner at the top
+    of the page mentioning when the site will be reset.
+    """
```
thibaudcolas, line 12: From what I understand this is something that would display at the top of all HTML responses regardless of the pages’ designs? This seems like too much to me – there are lots of places where we wouldn’t want such a disclaimer to start with, and even when we do, I don’t think we’d want to have a custom design for this that we’d need to keep supporting.

I think it’s very important the visual design of this disclaimer is as unobtrusive as possible, so we can get a sense of the visual design of Wagtail without distractions. My preference would be to either only have the disclaimer on the login screen in the existing `help_block` you’ve added, or login screen and in addition the Wagtail dashboard, also as a `help_block` that would naturally fit with Wagtail’s visual design.

bakerydemo/templates/base/basic_auth.html
```diff
@@ -0,0 +1,16 @@
...
+    article { display: block; text-align: center; margin: 0 auto; }
+    a { color: #dc8100; text-decoration: none; }
+    a:hover { color: #333; text-decoration: none; }
+</style>
+
+<article>
+    <h1>Authentication Required</h1>
+    <p>You must login to view this site</p>
+    <h3>Looking for Wagtail's demo instance? <a href="https://demo.wagtail.org">Click here</a>.</h3>
+</article>
```
thibaudcolas, line 16: Even if this is just a holding page it still needs to be valid HTML and be accessible. Let me know if you want some help with this.

Specific things that stand out:

- Needs to be valid HTML
- Font sizes should always be in relative units
- Keep link underlines
- Never use "click here" as link text

bakerydemo/templates/base/demo_banner.html
```diff
@@ -0,0 +1,7 @@
+{% load wagtailadmin_tags %}
+
+<link rel="stylesheet" href="{% versioned_static 'css/demo-banner.css' %}">
+
+<div id="demo-banner">
+    <h1>This is a demo site. Content created here will be reset at <a href="https://mytime.io/3am/UTC">3am UTC</a>.</h1>
```
thibaudcolas, line 6: On the copy – my preference would be for a much simpler "will be reset every day". I don’t think the specific time matters enough to warrant the maintenance effort of listing it here.

If we want to keep the time – then let’s try to use a service that doesn’t use a `.io` domain if we can.

bakerydemo/base/middleware.py
RealOrangeOne: Given quite how brittle the current implementation is, I agree. Given changes can only be made after being logged in anyway, that'll be enough for people to see.

bakerydemo/templates/base/demo_banner.html
RealOrangeOne: I think including a cadence is important, but agree that as it's not a time defined in the repo, we should be more vague. I've updated it to "daily" (and thus removed the link)

bakerydemo/base/management/commands/reset_admin_password.py
RealOrangeOne: Ended up being more complex than either of us thought, especially as usernames can change too: https://github.com/wagtail/bakerydemo/pull/369/commits/7b74ab114e58f0d85f189b629500a925be28211c

bakerydemo/templates/base/demo_banner.html
thibaudcolas: 👍 sounds cool to me

bakerydemo/settings/base.py
```diff
@@ -191,3 +437,91 @@
...
+            "level": "WARNING",
+            "propagate": False,
+        },
+    },
+}
+
+SHOW_DEMO_BANNER = os.getenv("SHOW_DEMO_BANNER", "false").lower() == "true"
+
+# Changing this? Be sure to update the initial data fixture
+DEFAULT_ADMIN_PASSWORD = "changeme"
```
RealOrangeOne, line 525: Question: Now we're preventing people from changing the default admin password - should we use a different password which encourages them _to_ change it?

Dockerfile
```diff
@@ -1,72 +1,68 @@
...
-        mime-support \
-        postgresql-client \
-        procps \
-        zlib1g \
-    " \
-    && seq 1 8 | xargs -I{} mkdir -p /usr/share/man/man{} \
-    && apt-get update && apt-get install -y --no-install-recommends $RUN_DEPS \
-    && rm -rf /var/lib/apt/lists/*
+# Install dependencies in a virtualenv
+ENV VIRTUAL_ENV=/venv
```
tomusher, line 8: Do we need a virtual env?

RealOrangeOne: As we're not using poetry here, technically no, but it makes the environment more like our other production applications

bakerydemo/settings/base.py
RealOrangeOne: `admin`?

Dockerfile
tomusher: Not a huge deal, just thinking that this is likely going to be used as a 'best practices' `Dockerfile` example for a lot of people starting out with Wagtail, and this adds a layer of complexity without adding much value.

bakerydemo/settings/base.py
zerolab: `password`?
basically anything that is not `changeme`

setup.cfg
```diff
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length=120
 exclude=venv,.venv,bakerydemo/settings/local.py
-ignore=E501,F405,W503
+ignore=E501,F405,W503,E203
```
gasman, line 4: I think this can be reverted now (the banner middleware was the bit of code breaking this rule, and I'm not sure why it had to...)

bakerydemo/templates/base/basic_auth.html
```diff
@@ -0,0 +1,16 @@
...
+    body { font: 20px sans-serif; color: #333; }
+    article { display: block; text-align: center; margin: 0 auto; }
+    a { color: #dc8100; text-decoration: none; }
+    a:hover { color: #333; text-decoration: none; }
+</style>
+
+<article>
+    <h1>Authentication Required</h1>
+    <p>You must login to view this site</p>
+    <h3>Looking for Wagtail's demo instance? <a href="https://demo.wagtail.org">Click here</a>.</h3>
```
gasman: Out of curiosity, what's the motivation for including basic auth support as part of this PR, and in what situation do you envisage people stumbling upon a password-protected bakerydemo instance while looking for the demo site? I feel like I've missed a memo about how we intend to use this :-)

bakerydemo/base/wagtail_hooks.py
```diff
@@ -74,3 +83,72 @@ class BakeryModelAdminGroup(ModelAdminGroup):
...
 modeladmin_register(BakeryModelAdminGroup)
+
+if settings.PREVENT_ADMIN_CREDENTIALS_CHANGE:
+
+    @hooks.register("after_edit_user")
+    def notify_rejected_admin_changes(request, user):
+        """
+        Notify the user when they attempt to change the admin user's password that it won't work.
+
+        See the `prevent_user_password_change` signal below for where the noop is done.
```
gasman, line 94:
```suggestion
        See the `prevent_admin_changes` signal below for where the noop is done.
```

bakerydemo/base/wagtail_hooks.py
```diff
@@ -74,3 +83,72 @@ class BakeryModelAdminGroup(ModelAdminGroup):
...
+
+        if instance.id != settings.DEFAULT_ADMIN_PK:
+            return
+
+        # Check that `set_password` was called
+        if instance._password:
+            instance.set_password(settings.DEFAULT_ADMIN_PASSWORD)
+
+        if instance.username != settings.DEFAULT_ADMIN_USERNAME:
+            instance.username = settings.DEFAULT_ADMIN_USERNAME
```
gasman, line 131: We should also block attempts to unset the `is_superuser` and `is_active` flags here.

bakerydemo/base/wagtail_hooks.py
```diff
@@ -74,3 +83,72 @@ class BakeryModelAdminGroup(ModelAdminGroup):
...
+
+    @hooks.register("before_delete_user")
```
gasman, line 133: Right now a user can also log in to the Django admin at `/django-admin/` and delete the admin account from there, which doesn't trigger this hook. I don't think there's any legitimate reason to access the Django admin on bakerydemo, so we should probably just remove it from urls.py entirely (or make it conditional on PREVENT_ADMIN_CREDENTIALS_CHANGE as the more cautious alternative).

gasman (changes requested):

bakerydemo/base/wagtail_hooks.py
tomusher: Spoke to @RealOrangeOne separately about this but I don't think we should allow *any* write access to the public demo - the chances this will be exploited to upload malicious or abusive material are too high.

Ideally we could allow logging in to the admin but not creating or editing any content, or we prevent admin access altogether and continue recommending Gitpod for users that want to try the admin.

gasman: IMO if we're preventing admin access then there's no real point in it existing at all - if someone wants to see what the front-end of a Wagtail website looks like, then there are plenty of real ones that are much better...

gasman:
Have just tested bringing up the updated code as both a standalone local instance and through https://github.com/wagtail/vagrant-wagtail-develop, and can confirm that both work fine (and the only immediately visible change is the default login credential message on the login screen).

bakerydemo/templates/base/basic_auth.html
RealOrangeOne: It's a great question I could have explained more. I included it here so that we can easily add protection to the current "latest" demo environment, moving it from a public instance to more of an internal (wagtail core etc) development tool. Putting basic auth adds a level of protection to it which we want, but in case people still have the link saved somewhere, this means they'll receive a link to the new demo instance. Note they only see the page if they fail the prompt a couple times (browser limitation, nothing we can do).

bakerydemo/base/wagtail_hooks.py
RealOrangeOne: I think adding a dedicated featureflag for the Django admin makes the most sense. That way it's still accessible for local setups / gitpod environments.

The debate about whether the instance should be writable at all is still being discussed internally.

bakerydemo/settings/base.py
RealOrangeOne: I've changed to `admin`, simply because it's the same as the username.

gasman (approved):

PR #371, opened 2022-11-24
Wagtail 4 considerations

nickmoreton:
Just a few Wagtail 4.0 Upgrade considerations.

PR #372, opened 2022-11-28
Additional panel definitions clean-up for Wagtail 4

thibaudcolas:
Follow-up to #371. This removes additional classes that no longer need adding to panels, and updates our InlinePanel definitions to use a different section heading and item label. Using different text isn’t required but helps a lot in understanding the interface.


nickmoreton (approved):
Looks good to me. Nice to know the `colx` classname can be removed as well.

thibaudcolas:
🙌 merging! Note for `colx`, they’re only superfluous in the cases where the 12-col grid is divided equally between all columns (so `6 6`, `4 4 4`, `3 3 3 3`, etc). In other cases they’re still useful.

PR #373, opened 2022-11-28
Replace FontAwesome usage with built-in icons for StreamField blocks

thibaudcolas:
StreamField only supports SVG icons since v2.15 (https://github.com/wagtail/wagtail/issues/8533). This switches the block icon definitions from FontAwesome with SVG icons built into Wagtail.

- `fa-quote-left` becomes `openquote`. 
- `fa-paragrpah` becomes`pilcrow`.
- `fa-s15` (a bathtub as I recall?) becomes `media` (same icon used for embeds in Wagtail within rich text.

I’ve updated the existing migrations directly, as the icon is unlikely to be something we’d actively use in a migration anyway.

### DIFF
--- a/bakerydemo/base/blocks.py
+++ b/bakerydemo/base/blocks.py
@@ -56,7 +56,7 @@ class BlockQuote(StructBlock):
     attribute_name = CharBlock(blank=True, required=False, label="e.g. Mary Berry")
 
     class Meta:
-        icon = "fa-quote-left"
+        icon = "openquote"
         template = "blocks/blockquote.html"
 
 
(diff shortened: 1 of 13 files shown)


zerolab:
worth dropping https://github.com/wagtail/bakerydemo/blob/main/requirements/base.txt#L4 while at it

thibaudcolas:
@zerolab this PR only changes icon definitions for StreamField blocks, there are other parts of the UI where we still use FontAwesome

zerolab:
I should stop commenting while ill 🤪 
for future ref - FA is used in the bread categories/people menu items, and in the footer

nickmoreton (approved):
Approved.

**Question**: does this mean the `wagtailfontawesome` dependency could also be removed?

thibaudcolas:
@nickmoreton ty! There are other parts of the UI that still use FontAwesome icons, so no. It’d be nice to remove it eventually.

PR #374, opened 2022-11-29
Bump django-storages version for Django 4 compatibility

cas--:
Fixes import error in storages/backends/s3boto3.py when deploying with Django 4.1

    ImportError: cannot import name 'force_text' from 'django.utils.encoding'

django-storages version 0.13 is first release with official Django 4.1 support.

Refs: https://github.com/jschneier/django-storages/blob/master/CHANGELOG.rst#113-2022-08-05


gasman:
Thanks @cas--! This fix is now included in #393 as a wider overhaul of the production config.

Issue #361, opened 2022-10-13
Docker setup - missing step


Big-Vi:
Hello,

I was trying to set up the WagTail Demo locally and using the specified commands did setup my environment. But it didn't allow me to log in to the backend(authentication error) until I ran the below command.

`docker-compose run app /venv/bin/python manage.py createsuperuser`

![Screen Shot 2022-10-14 at 9 46 31 AM](https://user-images.githubusercontent.com/30820063/195708144-3a85c515-99a6-42de-9563-5e672924ad9f.png)
![Screen Shot 2022-10-14 at 9 48 33 AM](https://user-images.githubusercontent.com/30820063/195708161-33165334-f721-422c-8542-621b829ee630.png)

Can I create a pull request by changing the instruction?

Cheers
V

zerolab:
Hey @Big-Vi,

Have you seen the

> Log into the admin with the credentials admin / changeme.

line in the README? did it not work for you?

Big-Vi:
The below credentials did not work for me.
username: admin
password: changeme

bc6really:
I agree with @cmuc24, thanks @Big-Vi.  I should also note that the hosted demo site is having a similar issue; when asked for the username and password, admin / changeme does not work.

gasman:
I think the issue with the hosted demo site is different - since there's a single database shared across all users, it's easy for anyone to casually change the default login (and they may well even think it's the right thing to do, because the password is "changeme" after all).

I'm not sure why the extra `createsuperuser` step would be necessary here, because the `admin` / `changeme` account is [hard-coded in the fixture along with the rest of the site data](https://github.com/wagtail/bakerydemo/blob/77d55ac019b1c6ea9d6c0647c5f22f7c42b8e9e6/bakerydemo/base/fixtures/bakerydemo.json#L1382-L1398) - if that account was missing, then the rest of the demo data would be missing too. We've previously had issues with the Docker setup due to the fact that the webserver setup and the database setup happen in parallel, so it's possible for the webserver to become available before the data is fully loaded - but again, if that was the problem here, you wouldn't see the bakery homepage either at that point (only the default "Welcome to your new Wagtail site" splash page).

mllabs:
I was also having trouble with logging in to either the docker container or the demo site.  The 'fix' does allow me to log in, but none of the data is there, so I figured it's probably due to the database not being up when the initial data load was running as mentioned by @gasman above.

Anyway - to fix both issues, I ran the command:

[...]

PR #375, opened 2022-12-05
Add missing migrate command

andylolz:
I think there’s a missing step in the docker-compose setup in the README. I think we need to run migrations before loading data.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -92,6 +92,7 @@ Run the following commands:
 git clone https://github.com/wagtail/bakerydemo.git
 cd bakerydemo
 docker-compose up --build -d
+docker-compose run app /venv/bin/python manage.py migrate
 docker-compose run app /venv/bin/python manage.py load_initial_data
 docker-compose up
 ```


gasman:
Not sure this is right - in https://github.com/wagtail/bakerydemo/issues/361 various people have reported the initial data (or at least the superuser account) being missing and having to re-run `./manage.py load_initial_data`, but nobody there has reported needing to run the migrate step. As per https://github.com/wagtail/bakerydemo/issues/361#issuecomment-1321232573, I suspect this is down to the database and webserver being initialised in parallel, meaning that these steps are being run before the database is ready.

andylolz:
You’re absolutely right! I wonder if I ran `load_initial_data` too quickly after running `docker-compose up --build -d`.

This is maybe a problem with running in detached mode – it’s not obvious that the migrations haven’t finished yet. But I don’t have a suggestion for how to resolve that.

Anyway – thanks for taking a look at this!

PR #376, opened 2022-12-22
Don't restart containers if stopped

tbrlpld:
When containers are stopped explicitly, it is very confusing to have them restart automatically. 
This is especially an issue in local dev where there is no needed for the db and redis containers to be running on every system/docker start.

This commit changes the restart policy to only restart when they have not been explicitly stopped. 
Explicitly stopping the container will not lead to a restart anymore.

See also: 
* https://docs.docker.com/compose/compose-file/#restart


thibaudcolas (approved):
Thank you, this has indeed been very confusing!

PR #382, opened 2023-01-03
Add missing end slash when generating tag URLs

gasman:
as per #381, but more Pythonic and without doing a useless `self.get_parent()` database hit on every iteration


thibaudcolas (approved):
Thank you @gasman, glad to see this looking much more sane.

PR #384, opened 2023-01-11
Productionize

RealOrangeOne:
This is an extension to #369, but without the public demo related features.

- Update the hosting config to one more "production-grade"
- Add a management command to reset the demo
- Prevent search engines indexing the site
- Add basic auth, with a custom template on failure


Dockerfile
```diff
@@ -1,72 +1,68 @@
-FROM python:3.9-slim
+# We use Debian images because they are considered more stable than the alpine
+# ones becase they use a different C compiler. Debian images also come with
+# all useful packages required for image manipulation out of the box. They
+# however weigh a lot, approx. up to 1.5GiB per built image.
+FROM python:3.9 as production
```
zerolab, line 5: my main gripe with this is that the image is quite a bit bigger now...

RealOrangeOne: Yeah it gets a **lot** bigger. There are some ways we can reduce the size of the image I've been meaning to try. Perhaps this is the time to try them... :thinking:

app.json
```diff
@@ -1,15 +0,0 @@
...
-  "env": {
-    "DJANGO_DEBUG": "off",
-    "DJANGO_SETTINGS_MODULE": "bakerydemo.settings.production",
-    "DJANGO_SECURE_SSL_REDIRECT": "on"
-  },
-  "scripts": {
-    "postdeploy": "django-admin migrate && django-admin load_initial_data && echo 'from wagtail.images.models import Rendition; Rendition.objects.all().delete()' | django-admin shell"
-  },
-  "addons": ["heroku-postgresql:hobby-dev"]
-}
```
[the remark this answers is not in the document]
RealOrangeOne: The reason to deploy to Heroku would be more for actually deploying it somewhere rather than just a "let me try this real quick". Especially given Heroku has removed its free tier, I think suggesting Heroku as a way to try Wagtail out isn't one that's going to get used too often. Gitpod works much better in those cases. There's little reason to deploy bakerydemo to Heroku.

Dockerfile
RealOrangeOne: I've made some changes internally to our image configuration. If they get reviewed and approved, we can update things here too. 45% improvement is definitely worthwhile.

tomusher:
This looks promising to me, however I do have a few concerns about the general direction this takes bakerydemo as a project.

While bakerydemo isn't necessarily a template, it is supposed to be a good starter example of a Wagtail site that is understandable and usable by beginners who are likely to be copying and pasting large parts of it in to their own projects, learning from the precedents set here.

Some of the changes here are positive in the interest of making it more consistent and deployable in the TBX world, but may be counterproductive in the pursuit of the 'starter' project goal. I've highlighted some specific examples in separate comments.

bakerydemo/static/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
tomusher, line 1: Leaving this in here feels like a potential footgun for a beginner learning from this project.

bakerydemo/templates/base.html
```diff
@@ -18,6 +18,9 @@
         <meta name="description" content="{% if page.search_description %}{{ page.search_description }}{% endif %}">
         <meta name="viewport" content="width=device-width, initial-scale=1">
 
+        {# Prevent the demo site from being indexed #}
+        <meta name="robots" content="noindex" />
```
tomusher, line 22: Same here, this template is likely to be copy and pasted around so I'm not sure this belongs here.

Dockerfile
tomusher: Do we gain anything by switching to the full-size image here? Appreciate we are willing to live with the tradeoffs in Torchbox projects, but as a recommended approach for beginners, is slim not preferrable?

Dockerfile
```diff
@@ -1,72 +1,68 @@
...
-        libpq5 \
-        mime-support \
-        postgresql-client \
-        procps \
-        zlib1g \
-    " \
-    && seq 1 8 | xargs -I{} mkdir -p /usr/share/man/man{} \
-    && apt-get update && apt-get install -y --no-install-recommends $RUN_DEPS \
-    && rm -rf /var/lib/apt/lists/*
+# Install dependencies in a virtualenv
```
tomusher, line 7: There are reasons we install dependencies in a virtualenv in our projects but this feels like another footgun for new users - is there any benefit from doing it here?

Dockerfile
```diff
@@ -1,72 +1,68 @@
...
+    DJANGO_SETTINGS_MODULE=bakerydemo.settings.production \
+    PORT=8000 \
+    WEB_CONCURRENCY=3 \
+    GUNICORN_CMD_ARGS="-c gunicorn-conf.py --max-requests 1200 --max-requests-jitter 50 --access-logfile - --timeout 25 --reload"
 
-# Base uWSGI configuration (you shouldn't need to change these):
-ENV UWSGI_VIRTUALENV=/venv UWSGI_HTTP=:8000 UWSGI_MASTER=1 UWSGI_HTTP_AUTO_CHUNKED=1 UWSGI_HTTP_KEEPALIVE=1 UWSGI_UID=1000 UWSGI_GID=2000 UWSGI_LAZY_APPS=1 UWSGI_WSGI_ENV_BEHAVIOR=holy
+# Make $BUILD_ENV available at runtime
+ARG BUILD_ENV
+ENV BUILD_ENV=${BUILD_ENV}
```
tomusher, line 41: Not sure if this is actually required except for our specific use case, so could probably remove it for simplicity. If you want to keep it, move the `ENV` instruction later in the file to prevent it causing layer cache issues.

app.json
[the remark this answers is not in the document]
tomusher: Not sure our messaging is clear on this then - we're shipping bakerydemo with a Dockerfile that mentions Heroku multiple times, and a `heroku.yml` file, but removing all the docs on how to use Heroku.

I think Heroku is still a valid and useful platform for those that want to quickly deploy bakerydemo for demoing to their team/for personal testing, and as everything else is tailored towards a Heroku deployment, why not keep it here?

bakerydemo/settings/base.py
```diff
@@ -20,12 +23,25 @@
 # See https://docs.djangoproject.com/en/3.2/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = "c6u0-9c!7nilj_ysatsda0(f@e_2mws2f!6m0n^o*4#*q#kzp)"
+# DJANGO_SECRET_KEY *should* be specified in the environment. If it's not, generate an ephemeral key.
```
tomusher, line 26: Environment variable mentioned here should be `SECRET_KEY`

bakerydemo/settings/base.py
```diff
@@ -66,18 +82,22 @@
     "taggit",
     "wagtailfontawesome",
     "debug_toolbar",
+    "django_extensions",
+    "simple_robots",
```
tomusher, line 86: Leaving this here also seems like a potential footgun - another dependency to install that beginners might not understand the use of.

bakerydemo/settings/base.py
```diff
@@ -114,12 +134,108 @@
...
+# Front-end cache
+# This configuration is used to allow purging pages from cache when they are
+# published.
+# These settings are usually used only on the production sites.
+# This is a configuration of the CDN/front-end cache that is used to cache the
+# production websites.
+# https://docs.wagtail.org/en/latest/reference/contrib/frontendcache.html
+# The backend can be configured to use an account-wide API key, or an API token with
+# restricted access.
+if (
```
tomusher, line 207: Cloudflare frontend cache seems a bit too specific to our own deployment to include here?

bakerydemo/settings/base.py
```diff
@@ -167,12 +283,88 @@
...
+    # https://docs.wagtail.io/en/stable/advanced_topics/deploying.html#cloud-storage
+    # Not having this setting may have consequences in losing files.
+    AWS_S3_FILE_OVERWRITE = False
+
+    # Default ACL for new files should be "private" - not accessible to the
+    # public. Images should be made available to public via the bucket policy,
+    # where the documents should use wagtail-storages.
+    AWS_DEFAULT_ACL = "private"
+
+    # We generally use this setting in the production to put the S3 bucket
```
tomusher, line 339: Perhaps we should change some of this language to be more instructive rather than descriptive of how 'we' do things.

tomusher:
Having read over this a bit further, I wonder if the best approach here is to separate out a 'Torchbox' deployment config for this in a separate repo, keeping the base config in here as simple as possible?

PR #385, opened 2023-01-27
Update imports that reference wagtail.core

gasman:
wagtail.core has been deprecated since 3.0, outputs warnings as of 4.2, and is due for removal in 5.0.

(Note that wagtailfontawesome still references `wagtail.core.hooks`, and since that has been unmaintained for a while we'll probably need to either switch away from that to an SVG solution, or fork / take ownership of wagtailfontawesome, before we can proceed with fully dropping wagtail.core)


zerolab:
imho, it is time to drop wagtailfontawesome

zerolab (approved):

laymonage (approved):

PR #387, opened 2023-01-28
Updated the readme file to include the information about the automatic conversion of line endings in windows machines

Asher-MS:


gasman:
Thanks @Asher-MS!
Since this is tutorial material, and possibly the first impression that people will have of Wagtail, it's important for us to keep these instructions as focused as possible, keep explanations minimal, and avoid giving too much prominence to "here are some ways things could go wrong"... see https://diataxis.fr/tutorials/#writing-a-good-tutorial.

Rather than adding an extra step here, could we change the `git clone` line to:

    git clone https://github.com/wagtail/bakerydemo.git --config core.autocrlf=input

as suggested in that article, so that it works consistently across all platforms. Then, if we still feel that an explanation is needed, add a sentence after that code block saying something like:

    (The `core.autocrlf` option is needed on Windows to ensure that line breaks are [preserved in the format that Docker requires](https://willi.am/blog/2016/08/11/docker-for-windows-dealing-with-windows-line-endings/).)

thibaudcolas:
Updated with guidance from Matthew, merging :)

PR #389, opened 2023-01-31
Add image rendition update step to Docker setup instructions

### LINKED ISSUE #388: Fixture images not loaded with initial Docker setup
I followed the steps for the initial Docker setup outline in the README. 

I do get a working site, but all images are missing. 

Not sure if there is some config that I missed.

tbrlpld:
Before, when following the Docker setup instructions, the images would not be served.

Running the update renditions step fixes this.

Fixes #388

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -98,6 +98,7 @@ Wait 10 seconds for the database setup to complete. Then run:
 
 ```bash
 docker compose run app /venv/bin/python manage.py load_initial_data
+docker compose run app /venv/bin/python manage.py wagtail_update_image_renditions
 ```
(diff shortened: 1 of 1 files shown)


gasman:
I think we ought to get to the bottom of why this is necessary, rather than just adding lines to the instructions as and when we encounter a problem. See #361, where numerous people were insistent that we needed to add a `migrate` / `load_initial_data` / `createsuperuser` step when actually they just needed to wait 10 seconds.

If this does turn out to be necessary, can this be made part of the setup script rather than another line for users to run?

gasman:
Closing as unable to replicate, as per #388.

PR #396, opened 2023-02-14
Manually change StreamField block identifiers for uniqueness

thibaudcolas:
Manually changes StreamField block identifiers to get around a bug in Wagtail where multiple blocks end up with the same identifier. I would recommend reviewing this with `git show HEAD --color-words` – the diffs in GitHub are much bigger than the actual change.

I have created https://github.com/wagtail/wagtail/issues/10073 to track the underlying bug in Wagtail, and in the meantime we can manually change the data here.


laymonage (approved):
Diff:

- `bakerydemo.json[238]["fields"]["body"]`, a `recipes.recipepage` of `pk` `81` aka "Hot Cross Bun".
  - See an `ingredients_list` block of `id` `21a44ce4-9aa6-459c-b654-d8d2217e8c8f`, there are two different instances with the same `id`.
  - The second instance's `id` has been replaced with `4ea44c12-9aa6-459c-b654-d8d2217e8c8f` in this PR.
- `bakerydemo.json[241]["fields"]["body"]`, a `recipes.recipepage` of `pk` `83` aka "Mincemeat Tart".
  - See a `steps_list` block of `id` `90b06737-b5c1-43be-9efd-d851b42df8bf`, there are two different instances with the same `id`.
  - The first instance's `id` has been replaced with `73b06709-b5c1-43be-9efd-d851b42df8bf` in this PR.

Can confirm these errors are now gone:

## Before

image
image

## After

image


---

The "Southern Cornbread" page does not have the issue, so 👍 

Thanks!

PR #397, opened 2023-02-14
Fix live preview for `BlogPage` and `RecipePage` authors

laymonage:
This was a regression in f381355cfe60cd56daef6eee49294dbcc63ebc72

I was curious why we had to do `[n.person for n in self.blog_person_relationship.all()]` before that commit, which would mean that we're querying the relationship table and then for each relationship record, we have to fetch the related `Person` object.

Thus, in that commit I changed it to query the person directly. However, in doing so it made the live preview not updating the Authors when you change the selections in the inline panel.

It turns out we needed to access the authors through `blog_person_relationship` so that we make use of django-modelcluster's features to allow reflecting the relationship changes without committing them to the database.

This reverts the logic to use the original logic, but I added a `select_related` to optimise the performance when accessing the method with real `blog_person_relationship` DB queries. Not sure if that's something django-modelcluster will keep under the hood, but at least it doesn't break...

### DIFF
--- a/bakerydemo/blog/models.py
+++ b/bakerydemo/blog/models.py
@@ -13,7 +13,6 @@ from wagtail.models import Orderable, Page
 from wagtail.search import index
 
 from bakerydemo.base.blocks import BaseStreamBlock
-from bakerydemo.base.models import Person
 
 
 class BlogPersonRelationship(Orderable, models.Model):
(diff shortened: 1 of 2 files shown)


thibaudcolas (approved):
Thank you @laymonage, all working as expected – both the previewing, and the specific logic for draft authors.

PR #400, opened 2023-02-20
Set python path

RealOrangeOne:
This is needed to allow the `django-admin` command to work


lb- (approved):

lb-:
Thanks, I think we also need this in the docker wagtail repo

PR #403, opened 2023-04-03
Updated docker-compose.yml

KingNonso:
Correctly map the volume  from your host to your container defined in [/code](https://github.com/wagtail/bakerydemo/blob/main/Dockerfile), which receives the latest code and uses it to update your application.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       context: .
       dockerfile: ./Dockerfile
     volumes:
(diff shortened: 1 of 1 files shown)


gasman:
@RealOrangeOne Please can you sanity-check this? Looks like #405 might be connected...

RealOrangeOne (approved):
Yep looks like this will indeed resolve #405 - nice catch!

PR #407, opened 2023-04-18
fixed setup with Docker

ACK1D:
App does not start in Docker, with these fixes everything returns to normal.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,4 +61,4 @@ VOLUME ["/code/bakerydemo/media/images/"]
 ENTRYPOINT ["/code/docker-entrypoint.sh"]
 
 # Start uWSGI
-CMD ["uwsgi", "/code/etc/uwsgi.ini"]
+CMD ["/venv/bin/uwsgi", "/code/etc/uwsgi.ini"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -61,4 +61,4 @@ VOLUME ["/code/bakerydemo/media/images/"]
 ENTRYPOINT ["/code/docker-entrypoint.sh"]
 
 # Start uWSGI
-CMD ["uwsgi", "/code/etc/uwsgi.ini"]
+CMD ["/venv/bin/uwsgi", "/code/etc/uwsgi.ini"]
```
RealOrangeOne, line 64: Issue: This change shouldn't be necessary, as `/venv/bin` is added to the `PATH` [here](https://github.com/wagtail/bakerydemo/blob/97624f71a96c12e12150cb497cc7b517ed5dd218/Dockerfile#L22)

RealOrangeOne (changes requested):

cnk:
Most of us do not appear to be having the issue @ACK1D was having and both parts of this PR have been reviewed and found to be things we do not want to include.

PR #413, opened 2023-05-15
Add features from Wagtail 5.0

laymonage:
- The `Person` and `FooterText` model are no longer registered with ModelAdmin. The menu items are registered through `SnippetViewSet` and `SnippetViewSetGroup`. The `BreadIngredient`, `BreadType`, and `Country` models are still registered with ModelAdmin for now to ease testing of ModelAdmin features while the contrib module still exists.
- A custom `AccessibilityItem` userbar item subclass is added to enable all non-experimental Axe rules.
- The `Person` model now extends `LockableMixin`, making it possible to lock/unlock `Person` instances in the CMS.
- The `Person` model now also extends `WorkflowMixin`. A migration has been added to also create the corresponding `WorkflowContentType` with an existing active `Workflow`. This allows you to submit a `Person` instance to the default "Moderators approval" workflow without any further configuration. A different workflow can be used by creating another workflow via Settings > Workflow and assigning it to the `Person` model (after unassigning it from the "Moderators approval" workflow).
- Not specifically from Wagtail 5.0, but the `FooterText` model has now been configured to be translatable to ease testing translation features with snippets. No actual setup has been made to use translated instances for the rendered user-facing pages.


gasman (approved):
LGTM!

Issue #415, opened 2023-06-15
Symbol before child pages in Breadcrumb path display


MikeMavrok:
---
Issue with Breadcrumb path display
---

## Description

**Issue:** The breadcrumb path in the bakerydemo site is showing an extra symbol (_) like underscore before some of the page names. For example, "Home > _Breads" instead of "Home > Breads".

This issue is observed in the Wagtail/bakerydemo repository.

## Steps to Reproduce

1. Visit the bakerydemo site.
2. Navigate to any page that has child pages.
3. Observe the breadcrumb navigation at the top of the page.

## Expected Behavior

The breadcrumb should display the correct path without any extra symbols.

## Actual Behavior

The breadcrumb path contains an extra symbol (_) before certain page names.

## Screenshots

![image](https://github.com/wagtail/bakerydemo/assets/93074951/b37462ac-0ac8-43d1-8f8d-d643c65f05ac)

## Environment

- Wagtail version: 5.0
- Docker version 23.0.5
- Browser: Chrome
- Operating System: Windows 10

## Additional Information

The error is probably located in [breadcrumbs.html](https://github.com/wagtail/bakerydemo/blob/046eb8ecf866b88b6e4f402a66e04eefea93f504/bakerydemo/templates/tags/breadcrumbs.html#L4)

laymonage:
The "symbol" is actually an underline from CSS. This bug is a result of the `text-decoration: underline;` for the `.breadcrumb` selector:

https://github.com/wagtail/bakerydemo/blob/046eb8ecf866b88b6e4f402a66e04eefea93f504/bakerydemo/static/css/main.css#L836-L845

The `.breadcrumb` selector covers the whole breadcrumb. We only want the underline to be applied on the links, and in fact we already do that here:

https://github.com/wagtail/bakerydemo/blob/046eb8ecf866b88b6e4f402a66e04eefea93f504/bakerydemo/static/css/main.css#L861-L865

So the correct fix should be to remove the `text-decoration` in `.breadcrumb`:

![image](https://github.com/wagtail/bakerydemo/assets/6379424/4d8ed8b6-f39a-4861-9e28-5cd58e8d6387)

lb-:
Fixed via https://github.com/wagtail/bakerydemo/pull/416

PR #416, opened 2023-06-15
Remove unnecessary underline from breadcrumb (Issue #415)

gzark1:
I removed a line in `bakerydemo/static/css/main.css` which was causing the page names in breadcrumb to have an unnecessary line, as suggested by @laymonage in #415 

**Before:**
image

**After:**
image


lb-:
Thanks @gzark1 - appreciate it.

PR #418, opened 2023-06-16
Fix iframe dimensions in rich text. Fix #417

### LINKED ISSUE #417: Problem with iframe dimensions in paragraph blocks in Streamfields.
**Description:**

The dimensions of iframes within the .block-paragraph_block containers are excessively small, leading to suboptimal visibility and user experience.

I discovered that the current CSS selector for .block-embed_block iframe is not adequately capturing iframes within paragraph blocks. Consequently, the iframe dimensions remain unchanged, resulting in the undesirably small display.

**Steps to Reproduce:**

Navigate to the following page: http://localhost:8000/blog/desserts-benefits/
Scroll down to the section where there is a YouTube video embedded within a paragraph.
Observe that the YouTube video appears to be smaller than expected within the paragraph container.

**Expected Behavior:**

The YouTube video embedded within the paragraph block should have appropriate dimensions and aspect ratio, ensuring optimal visibility and readability.

**Actual Behavior:**

The .block-embed_block iframe CSS rule, which sets the aspect ratio to 16:9 and adjusts the width and height to 100%, does not cover the case when an iframe is inside a paragraph block. As a result, the YouTube video within the paragraph block appears smaller than intended.

**Proposed Solution:**

I will open a pull request to implement the suggested CSS selector adjustment. The pull request will contain the necessary changes to accommodate iframes within paragraph blocks.

Additionally, I welcome any suggestions or alternative approaches from the reviewers to tackle this issue effectively.

**Screenshots** 

image

image

gzark1:
Fixes #147 

**Description:**

This pull request addresses the issue with iframe dimensions within the .block-paragraph_block containers. The existing CSS selector for .block-embed_block iframe does not cover iframes within paragraph blocks. To resolve this, the proposed solution includes an additional CSS rule to adjust iframe dimensions within paragraph blocks. 

**Changes Made:**

Added CSS selector to target iframes within .block-paragraph_block container:

image

Screenshot_2023-06-16_15-46-30

**Note to Reviewers:**

Your feedback and suggestions regarding the proposed solution are highly appreciated. Please review the changes and let me know if there are any further improvements or alternative approaches to consider.

### DIFF
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -405,6 +405,13 @@ blockquote .attribute-name {
 }
 
 /* stylelint-disable-next-line selector-class-pattern */
+.block-paragraph_block iframe {
+  aspect-ratio: 16 / 9 ;
+  width: 100%;
+  height: 100%;
+}
+
+/* stylelint-disable-next-line selector-class-pattern */
 :is(.block-table_block, .block-typed_table_block) :is(table, tr, td, th) {
   border: 1px solid var(--dark);
   padding: 0.25em;
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -406,7 +406,7 @@ blockquote .attribute-name {
 
 /* stylelint-disable-next-line selector-class-pattern */
 .block-paragraph_block iframe {
-  aspect-ratio: 16 / 9 ;
+  aspect-ratio: 16 / 9;
   width: 100%;
   height: 100%;
 }


lb-:
This won't work in IE 11 or Safari 13 & below. Probably not something we are aiming to support for this demo code though.

gzark1:
Would you suggest an alternative to this PR? Or keep it as is?

lb-:
I think it's good to go, just seeing if anyone else has thoughts before I merge. 

I'll check back in next week.

gzark1:
Okay, thank you for checking.

thibaudcolas (approved):
Looks excellent, thank you @gzark1!

Issue #395, opened 2023-02-07
Add Site Setting to the demo


lb-:
Site Settings are a key part of many Wagtail sites https://docs.wagtail.org/en/stable/reference/contrib/settings.html

It would be good to have an example in this demo site of a Site setting, ideally one that is translatable.

An example use case is some footer content that uses `BaseSiteSetting` (that can be translated).

Additionally, it would be great to have something that uses the `BaseGenericSetting` such as a logo.

See also https://github.com/wagtail/bakerydemo/issues/175

Imshk:
I would to like to help on this issue , please provide me more information so I can solve it .

Hanoon02:
Hi, I have made a rough sketch of the steps needed to complete this task. Points taken from https://docs.wagtail.org/en/stable/reference/contrib/settings.html. If its valid can I do a PR and add these changes?

We create a new app within the bakerydemo project using `python manage.py startapp mysettings`
Add it to the `INSTALLED_APPS` setting - 
```
INSTALLED_APPS = [
 ...
    'mysettings',
 ...
]
```
Now, create a `models.py` file in the `mysettings` app directory and import the `BaseSiteSettings` model from `wagtail.contrib.settings.models`
```
from wagtail.contrib.settings.models import BaseSiteSettings

class MySiteSettings(BaseSiteSettings):
    footer_text = models.CharField(
        verbose_name='Footer text',
        max_length=255,
        null=True,
        blank=True,
    )

    panels = BaseSiteSettings.panels + [
        FieldPanel('footer_text'),
    ]

    class Meta:
        verbose_name = 'Site settings'
        verbose_name_plural = 'Site settings'
```

Here, I've made MySiteSettings model that extends the `BaseSiteSettings` model and adds a `footer_text` field. It also have new `panels` attribute that includes the `footer_text` field in the admin interface.

Now, to use the new Site Settings, register it with Wagtail `wagtail_hooks.py` file
```
from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
from .models import MySiteSettings

class MySiteSettingsAdmin(ModelAdmin):
    model = MySiteSettings
    menu_label = 'Site settings'

modeladmin_register(MySiteSettingsAdmin)
```
I have extended the ModelAdmin class and set the model attribute to the new setting

Finally run the `make migration` codes as usual

[...]

PR #419, opened 2023-06-21
Add Generic and Site settings models

Morsey187:
Resolves issue: [395](https://github.com/wagtail/bakerydemo/issues/395)

Adds:
- `SiteSettings` extending from Wagtail's `BaseSiteSetting` class
    - New field `title_suffix` which is used to replace static text in the base.html's `title_suffix` block
- `GenericSettings` extending from Wagtail's `BaseGenericSetting` class
    - New fields`github_url`, `twitter_url`, `organisation_url` which are now used to conditionally display the footer icons and populate their href attributes.
- Fixture update for `bakerydemo.json` i.e. `load_initial_data`


bakerydemo/base/fixtures/bakerydemo.json
```diff
@@ -1624,6 +1600,24 @@
       "user_permissions": []
     }
   },
+  {
```
Morsey187, line 1603: Ah sorry missed this

Morsey187: Will push an update for this a bit later

bakerydemo/base/fixtures/bakerydemo.json
```diff
@@ -1444,6 +1365,13 @@
       "workflow": 1
     }
   },
+  {
+    "model": "wagtailcore.workflowcontenttype",
```
Morsey187, line 1369: From what I can tell this is a new model from a minor wagtail release that just hasn't been included until now

bakerydemo/base/fixtures/bakerydemo.json
```diff
@@ -3587,6 +3515,22 @@
       "file_hash": ""
     }
   },
+  {
+    "model": "wagtailusers.userprofile",
```
Morsey187, line 3519: Not sure why this was added, doesn't seem to break anything but can't figure out how I managed to add it in either.

thibaudcolas (approved):
All looking great @Morsey187! The fixtures changes surprise me a bit but yeah we don’t update the fixtures that often so I’ll probs just split the unrelated changes to a different commit and move on.

bakerydemo/base/fixtures/bakerydemo.json
thibaudcolas: It seems that for some reason we didn’t have a Wagtail user profile instance in the fixtures until now. Perhaps this gets created only when accessed the first time – which we now do on every page as part of determining which theme to show the UI in.

lb-:
Thanks @Morsey187 and @thibaudcolas - this is a really helpful addition

PR #420, opened 2023-06-24
create apps.py and modify base.py in order to set deafult lazy loadin…

MikeMavrok:
### Description

The purpose of this pull request is to ensure that all images in the Wagtail Bakerydemo project load lazily by default. This change demonstrates how easy it is to enhance the performance of the site.

### Changes Made

- Created `bakerydemo/base/apps.py` 
- Inside apps.py created a custom `CustomImagesAppConfig` class that extends the `WagtailImagesAppConfig` provided by Wagtail.
- Added the `default_attrs` attribute to the `CustomImagesAppConfig` class, specifying additional default attributes for all images.
- Set the `default_attrs` to include `"decoding": "async"` and `"loading": "lazy"`, improving the loading and decoding behavior of images.
- Removed `"wagtail.images"` from the `INSTALLED_APPS` list in the `settings/base.py` file.
- In the `settings/base.py` file, updated the `INSTALLED_APPS` list to include the path to the custom `CustomImagesAppConfig` for images.

### Motivation

By setting the default attributes for images to include `"loading": "lazy"`, the site benefits from lazy loading, where images are loaded only when they enter the viewport. This approach improves the initial page load time and overall performance of the site, especially for pages with numerous images.

### Additional Notes

- The custom configuration demonstrates the flexibility of Wagtail and its ability to extend and customize functionalities.
- This pull request showcases a simple yet effective technique for optimizing site performance.

This change aligns with best practices for web performance and highlights how easy it is to create a more performant site using Wagtail.

Please review the changes and let me know if there are any further modifications required.


bakerydemo/settings/base.py
```diff
@@ -48,7 +48,7 @@
     "wagtail.users",
     "wagtail.snippets",
     "wagtail.documents",
-    "wagtail.images",
+    "bakerydemo.base.apps.CustomImagesAppConfig",
```
lb-, line 51: Can't we just add a config to the base app? Do we need this additional import here?

MikeMavrok: Thanks for your answer! I think this import is necessary for the  `INSTALLED_APPS` in order to read the path to `CustomImagesAppConfig`. What is your suggestion?

lb-: Ahh I see. This replaces the images app. 

In that case I recommend we add a new app, `bakerydemo/images` to house this standalone config.

MikeMavrok: So if I am not mistaken, do you mean to create a new file `bakerydemo/images.py`  and copy the context of `bakerydemo/base/apps.py` ?

lb-: Sorry, it probably wasn't clear. 

My concern with the current approach is that it's a bit confusing, the base app contains the app config for the images app override. 

Instead, create a new app and put the apps.py in there. Django will automatically pick up the apps.py

Which means the line above will simply be `'bakerydemo.images'`.

You can create a new app manually (just a folder) or with the Django command. Something like `python manage.py startapp polls`

Docs for how apps.py / app Configs are loaded. 

https://docs.djangoproject.com/en/4.2/ref/applications/#configuring-applications

lb-: One other approach, we could just have a `apps.py` added at the bakerydemo sub-folder root. That may be even simpler.

MikeMavrok: Thank you for your clear instructions. It was actually more complex than needed. I updated this change in my last [commit](https://github.com/wagtail/bakerydemo/pull/420/commits/d43e31537872df8e8ef9b9a31a7f2bcb296c3bfe)

MikeMavrok:
I conducted some tests based on my previous [commit](https://github.com/wagtail/bakerydemo/pull/420/commits/d43e31537872df8e8ef9b9a31a7f2bcb296c3bfe) where I reverted to including "wagtail.images" in the `INSTALLED_APPS`. The purpose of these tests was to verify whether lazy loading was active or not, specifically focusing on the DOMContentLoaded time using Chrome Dev Tools and the "Slow 3G" network conditions.

Here are the results:

1. The first screenshot displays the DOMContentLoaded time when the `loading="lazy"` attribute was present in the `<figure class="picture-card__image">` element located in `templates/tags/gallery.html`. The DOMContentLoaded time was recorded as 8.74 seconds.
   
screenshot1


2. The second screenshot illustrates the DOMContentLoaded time when the `loading="lazy"` attribute was removed and "wagtail.images" was included in the `INSTALLED_APPS` list of `settings/base.py`. The DOMContentLoaded time significantly increased to 16.12 seconds.
   
screenshot2


3. The third screenshot represents the DOMContentLoaded time when the `loading="lazy"` attribute was removed and "bakerydemo.apps.CustomImagesAppConfig" was included in the `INSTALLED_APPS` list of `settings/base.py`. The DOMContentLoaded time improved and returned to 8.73 seconds.
   
screenshot3

Based on these observations, it appears that "wagtail.images" overrides the CustomImagesAppConfig configuration. This led me to include the import statement "bakerydemo.apps.CustomImagesAppConfig" in the INSTALLED_APPS list.

lb-:
Change is looking good, myself or someone else will do a manual review before merging. 

Thanks for working through this.

thibaudcolas (changes requested):
👋  thanks for the suggestion @MikeMavrok. I like the idea of using `loading="lazy"` on this site but I’m not convinced this is the right approach, or there is more to do.

We have a lot of hero images on this site, which we wouldn’t want to be lazy-loaded. We’d need to either:

- Retain your proposed approach and manually set hero images to `loading="eager" decoding="sync"` – if this is supported (not super clear to me)
- Or remove the app config and add `loading="lazy" decoding="async"` to non-hero images manually

thibaudcolas:
I’m going to close this now, as there’s been no further activity and as mentioned above I don’t think this is a good idea due to the amount of hero images on this site. We’d just be introducing a lot of layout shift for no reason, when we could instead apply lazy-loading only to the images that would actually benefit from it.

PR #422, opened 2023-06-26
Add new custom block "DetailsBlock" with expanding answers and demonstrate it in a BreadPage and in a new FAQ page.

gzark1:
@lb- suggested adding a `DetailsBlock` to showcase a custom block with its own HTML template. The `DetailsBlock` includes three fields: `summary` (CharField), `content` (RichTextField), and `open` (BooleanField, defaulting to open). This block is ideal for creating an FAQ page with expanding answers and allows for the use of semantic and modern HTML.

**Changes Made:**

1. Created a `DetailsBlock` class inheriting from `StructBlock` with the specified fields.
2. Added the `DetailsBlock` to the StreamField used by other pages for their content, providing them with the option to include a details block.

image

3. Created a template, `details_block.html`, for rendering the `DetailsBlock`.
4. Added CSS selectors for the `DetailsBlock` summary and content in the main CSS file.
5. Demonstrated the usage of the `DetailsBlock` in a bread page. 

image

6. Implemented a standalone StreamField with only the `DetailsBlock` option, showcased in a new `FAQ` page of the `StandardPage` type.

image

image

7. Created sample data for both the bread page and the FAQ page, stored in `bakerydemo/base/fixtures/bakerydemo.json`.

8. Added the `FAQ` page to the navigation menu.

image

PR #425, opened 2023-06-28
Improved: Add new custom block "DetailsBlock" with expanding answers and demonstrate it in a BreadPage and in a new FAQ page

gzark1:
This PR is an improvement of #422. I deleted the DetailsBlock option from `BaseStreamBlock`, in order not to complicate things. I utilized `DetailsStreamBlock` in both `BreadPage` (and demonstrated it in "Anadama" bread) and `StandardPage` (demonstrated it in "FAQ"). I stored the data i created in `bakerydemo/base/fixtures/bakerydemo.json`, keeping only the necessary changes.


lb-:
Looking much better @gzark1 

Do you mind reviewing the CI error, might need to make some more tweaks to the fixtures it seems.

cnk:
I have fixed the merge conflicts in the bakerydemo.json fixture. Now we are in a position to evaluate whether or not to accept this contribution.
Screenshot 2023-10-08 at 12 45 36 AM
Screenshot 2023-10-08 at 12 46 02 AM

gzark1:
Thank you @cnk for addressing the merge conflicts. Please feel free to provide any further guidance or suggestions, and I'll promptly address them.

bakerydemo/base/blocks.py
```diff
@@ -60,6 +61,22 @@ class Meta:
...
 
+class DetailsBlock(StructBlock):
+    """
+    Custom `StructBlock` for creating a details block with summary, content, and open fields.
+    """
+
+    summary = CharBlock(required=True)
+    content = RichTextBlock(required=True)
+    open = BooleanBlock(
+        required=False, default=True, label="Open", help_text="Open by default"
```
thibaudcolas, line 72: I’m not too clear why we are using a `<details>` element in HTML if it’s to have those blocks opened by default. Could we default to "False" instead?

I’d also suggest removing the label if it just repeats the field identifier, and writing a more descriptive help text or removing it (stating the default value isn’t too helpful).

bakerydemo/base/blocks.py
```diff
@@ -77,3 +94,12 @@ class BaseStreamBlock(StreamBlock):
         icon="media",
         template="blocks/embed_block.html",
     )
+
+
+# StreamBlocks for Details
+class DetailsStreamBlock(StreamBlock):
```
thibaudcolas, line 100: I’m really surprised we’re not adding this to `BaseStreamBlock` – this feels like a clear missed opportunity to demonstrate Wagtail’s flexibility, by allowing series of "details" content to be placed at arbitrary points?

thibaudcolas (changes requested):
I think this is great – but I would say this ought to be in the `body` field, so we can demonstrate Wagtail’s support for flexible page content.

lb-:
@gzark1  are you able to rebase and review the comments above?

Issue #423, opened 2023-06-27
CSP testing setup


thibaudcolas:
Part of [#1288 CSP compatibility issues](https://github.com/wagtail/wagtail/issues/1288). We need a basic opt-in setup on the bakerydemo so we can try out different degrees of CSP strictness (see also #232). This will simplify the process for contributors to make improvements to CSP support.

Here are my thoughts on the setup but this is very much not something I’m knowledgeable about so will need guidance from others:

- This could be set up with the `Content-Security-Policy-Report-Only` header for now, so we can iterate on CSP support without disrupting demo site usage.
- Either with `django-csp` or from scratch
- Ideally in a way that’s configurable easily so we can try different levels of strictness (`unsafe-eval`, `unsafe-inline`, `script-dynamic`)

It’d be great for the default configuration to be one we could recommend for Wagtail sites, but if that seems too tricky, any configuration will be better than nothing.

thibaudcolas:
For reference, this is what Torchbox projects do as a baseline, with `django-csp`:

```python
# Content Security policy settings
# http://django-csp.readthedocs.io/en/latest/configuration.html
if "CSP_DEFAULT_SRC" in env:
    MIDDLEWARE.append("csp.middleware.CSPMiddleware")

    # The “special” source values of 'self', 'unsafe-inline', 'unsafe-eval', and 'none' must be quoted!
    # e.g.: CSP_DEFAULT_SRC = "'self'" Without quotes they will not work as intended.

    CSP_DEFAULT_SRC = env.get("CSP_DEFAULT_SRC").split(",")
    if "CSP_SCRIPT_SRC" in env:
        CSP_SCRIPT_SRC = env.get("CSP_SCRIPT_SRC").split(",")
    if "CSP_STYLE_SRC" in env:
        CSP_STYLE_SRC = env.get("CSP_STYLE_SRC").split(",")
    if "CSP_IMG_SRC" in env:
        CSP_IMG_SRC = env.get("CSP_IMG_SRC").split(",")
    if "CSP_CONNECT_SRC" in env:
        CSP_CONNECT_SRC = env.get("CSP_CONNECT_SRC").split(",")
    if "CSP_FONT_SRC" in env:
        CSP_FONT_SRC = env.get("CSP_FONT_SRC").split(",")
    if "CSP_BASE_URI" in env:
        CSP_BASE_URI = env.get("CSP_BASE_URI").split(",")
    if "CSP_OBJECT_SRC" in env:
[...]

PR #426, opened 2023-06-30
Add Content Security Policy testing setup

Stormheg:
To be used to test Wagtail's compatibility with [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy).

By default, CSP is only enabled when the `CSP_DEFAULT_SRC` environment variable is set. This environment variable functions as a feature flag because we don't want to frustrate normal users with a broken website or a lot of CSP violation errors in the browser console.

A default configuration for `CSP_DEFAULT_SRC` can be loaded by sourcing the `csp.env` file added by this PR. Other CSP policy values can be influenced through environment variables as well.

Related:
- https://github.com/wagtail/wagtail/issues/1288
- https://github.com/wagtail/bakerydemo/issues/423

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@
 # If the variables defined here are loaded into the environment, CSP will be enabled.
 
 # Careful about the quoting of directives! It is easy to break.
-export CSP_DEFAULT_SRC="'self'"
+# CSP_DEFAULT_SRC="'self'"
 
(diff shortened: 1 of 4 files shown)


thibaudcolas (approved):
Looking great @Stormheg!

PR #427, opened 2023-06-30
Remove the need for inline styles for hero images

thibaudcolas:
Follow-up to #232. I’ve switched to `image` tags without alt text to preserve the same behavior as before. As far as I can see the pages should look 1:1 identical, and this also works when the page doesn’t have an image.

Along the way I also noted we have lots of useless `{% image %}` tag usage. I’ve removed those.

### DIFF
--- a/bakerydemo/static/css/main.css
+++ b/bakerydemo/static/css/main.css
@@ -280,11 +280,17 @@ figure img {
 
 /* Hero image area */
 .hero {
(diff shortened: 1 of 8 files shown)


Stormheg (approved):
LGTM!

This resolves the Content Security Policy issue we just discussed by removing the non-compliant inline `style` attribute.

PR #429, opened 2023-06-30
Use a local copy of font awesome instead of a CDN

Stormheg:
Loading css and font files from an external resource violates strict Content-Security-Policy. We don't want to make an exception for a CDN because we want to maintain a *very* strict policy.

I've removed the .ttf, .eot and .woff font variants as they are for compatibility with browsers we don't support. References to these files have been removed from the minified css file provided by Font Awesome.

Files downloaded from: https://fontawesome.com/v4/get-started/


thibaudcolas (approved):
🌈 CSP compatibility, and I can run the bakerydemo on the plane. Dreamy!

lb-:
This is great. Thanks all.

Issue #338, opened 2022-06-01
Whitenoise is failing to find font files during Docker build


vossisboss:
I'm getting the following error when attempting a local installation in Docker:

```

> [8/9] RUN DATABASE_URL=postgres://none REDIS_URL=none /venv/bin/python manage.py collectstatic --noinput:                                              
#12 0.238 /venv/lib/python3.9/site-packages/dotenv.py:56: UserWarning: Not reading /code/.env - it doesn't exist.                                         
#12 0.238   warnings.warn("Not reading {0} - it doesn't exist.".format(dotenv))                                                                           
#12 0.865 Post-processing 'css/bootstrap.min.css' failed!                                                                                                 
#12 0.865                                                                                                                                                 
#12 0.865 Traceback (most recent call last):
#12 0.865   File "/code/manage.py", line 15, in <module>
#12 0.865 WARNING: DJANGO_SECRET_KEY not found in os.environ. Generating ephemeral SECRET_KEY.
#12 0.865     execute_from_command_line(sys.argv)
#12 0.865   File "/venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
#12 0.865     utility.execute()
#12 0.865   File "/venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 440, in execute
#12 0.866     self.fetch_command(subcommand).run_from_argv(self.argv)
#12 0.866   File "/venv/lib/python3.9/site-packages/django/core/management/base.py", line 414, in run_from_argv
#12 0.866     self.execute(*args, **cmd_options)
#12 0.866   File "/venv/lib/python3.9/site-packages/django/core/management/base.py", line 460, in execute
#12 0.866     output = self.handle(*args, **options)
#12 0.866   File "/venv/lib/python3.9/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 209, in handle
#12 0.866     collected = self.collect()
#12 0.866   File "/venv/lib/python3.9/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 154, in collect
#12 0.867     raise processed
#12 0.867 whitenoise.storage.MissingFileError: The file 'fonts/glyphicons-halflings-regular.eot' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0xffff8cf5dc70>.
#12 0.867 
#12 0.867 The CSS file 'css/bootstrap.min.css' references a file which could not be found:
[...]

PR #430, opened 2023-06-30
Remove old glyphicons font files and references

Stormheg:
The .eot, .ttf, .svg and .woff files exist for compatibility with browsers we no longer support.

It was impossible to update the source map file for bootstrap by hand so I have removed it.

This time, removing these file has been done right and https://github.com/wagtail/bakerydemo/issues/338 should not resurface. I checked by running `collectstatic` locally with `DJANGO_SETTINGS_MODULE=bakerydemo.settings.production` and no errors were reported by Whitenoise.


bakerydemo/static/css/bootstrap.min.css
```diff
@@ -2,5 +2,5 @@
...
+/*# sourceMappingURL=bootstrap.min.css.map */
```
thibaudcolas, line 6: It feels safer for us to remove this too if we remove the `.map` file?

Stormheg: Already done!

thibaudcolas (approved):
Great! No issues running `collectstatic`.

PR #431, opened 2023-06-30
Remove unneeded font files

thibaudcolas:
Expands upon #429 and #430 by removing more font files:

- Glyphicons was only used in one template which itself wasn’t even used.
- Font Awesome was only used for three footer icons. We can use FontAwesome SVGs instead.
- Open Sans can be replaced by a (usually) sans-serif system font stack, rather than needing 10 weights.


lb- (approved):
Looks good - let's do it.

PR #434, opened 2023-07-21
Clear cache when resetting demo

RealOrangeOne:
The Wagtail demo environment has a redis cache, which doesn't get cleared during the reset. This can result in stale objects in the cache and errors during page loads.


zerolab (approved):
Works for me.

PR #435, opened 2023-07-26
Use separate wagtail-modeladmin package

laymonage:
Not sure if we want this in bakerydemo, as we'd probably want to fully remove modeladmin from bakerydemo once the contrib module is out of the Wagtail repository anyway.

Putting this here just for demonstrating the replacement of `wagtail.contrib.modeladmin` with `wagtail_modeladmin`.


thibaudcolas:
Taking this forward with #439, which has a slightly better order of INSTALLED_APPS and also updates fixtures.

Issue #175, opened 2017-11-24
Make the demo more suitable for Wagtail testing


thibaudcolas:
I've been trying to set up automated testing for Wagtail using the bakerydemo. Since it's also used as the default development project by Wagtail contributors, I think it would make sense for it to be the go-to project for automated and manual testing.

Overall, the bakerydemo does much more than the previous wagtaildemo so it's already a great leap forward. Nonetheless, there are many things that are hard to test at the moment, of which I made a list below. There is one general theme: more features should be used in the demo "out of the box", so that there is the least setup / content loading required when we need to test a given feature.

Here is the list:

- [x] Wagtail styleguide – should be enabled in the `INSTALLED_APPS`.
- ~~[ ] Making sessions last a very long time (infinite!).~~
- [x] Have different accounts with different roles, in different groups.
- ~~[ ] Have an account with Gravatar.~~
- [x] Have promoted search results.
- [x] Have redirects.
- [ ] Have documents.
- [x] Have content/media in collections.
- [x] Have focal area used on some images.
- [x] Enable the image link generator (I couldn't find it, might have missed it)
- [ ] Have revisions for some pages
- [ ] Have pages in draft, live + draft status.
- [x] Add rich text using all available formats: bold, italic, p, h2, h3, h4, UL, OL, hr, embed, docs, image, link
- [x] Add Streamfield instances, and content, using more complex features (#168).

Before we go creating fixtures, I think it would be good to discuss which of those things would also be useful for people experimenting with Wagtail, and which are clearly catering for testing only and should be "hidden" one way or another from the normal bakerydemo experience.

heymonkeyriot:
We had quite a few conversations along these lines when first setting up the demo. The general consensus was that we should try and keep things as simple as possible for first time users, whilst giving enough complexity for people new to Wagtail and/or Django to starting getting a sense of how things work. I've no data to back it up with but personally seeing people engage with the demo for the first time I feel like we've got that balance of complexity right.

That said, it definitely could work better as a developer resource. Could one solution be to point to a specific branch for developer installs - much as we currently have where [vagrant-wagtail-develop](https://github.com/wagtail/vagrant-wagtail-develop) is currently pointing to the Wagtail 2.0 branch?

Agree on all counts with what we need in here. The only one I think I disagree with is:
> Making sessions last a very long time (infinite!).
Given how critical login is to the general health of the CMS I think it's important that we be forced through it regularly.

thibaudcolas:
[...]

PR #436, opened 2023-08-10
Add more demo data & features

thibaudcolas:
Solves part of #175, particularly things we’d want to have for demos & audits or common features. This might be easier to review com

- Promoted search results. Includes a basic implementation of the feature, and demo data.
- Demo users, to simplify switching between languages and groups, as well as checking what an inactive user looks like
- Demo redirects, based on real examples from the project (as part of renaming pages where the slug was incorrect, and demonstrating common typos)
- Wagtail’s dynamic image serve view. Bit of a niche feature but useful to be able to demo it when needed
- Focal points

Compared to #175, still missing are pages in different states, revisions, and documents


bakerydemo/base/fixtures/bakerydemo.json
```diff
@@ -1552,6 +1671,96 @@
...
+      "first_name": "Editor",
+      "last_name": "User",
+      "email": "editor@example.com",
+      "is_staff": false,
+      "is_active": true,
+      "date_joined": "2019-02-17T08:01:33.700Z",
+      "groups": [["Editors"]],
+      "user_permissions": []
+    }
+  },
```
Stormheg, line 1691: **Suggestion:** we should document that these users exist and why.

Stormheg (approved):
I'm happy with this!

I've added a section to the readme documenting the different users. Does this look okay to you @thibaudcolas?

Feel free to merge.

bakerydemo/base/fixtures/bakerydemo.json
laymonage: I had a commit in my fork that does a similar thing here: laymonage/bakerydemo@f42d7d1491941fa5313c8475ea7e0074b24442dc

In my branch, I used these users for benchmarking the SQL queries (different permissions need different queries). They're also useful for testing workflows.

Definitely would be nice to document these.

To add, in my case I used names that are distinct from the group name, e.g. "Eddy Thor" Instead of "Editor User" just in case we need the distinction when we want to write tests.

Stormheg: @laymonage are you suggesting that you want to rename the users? The current names feel adequate to me.

bakerydemo/base/fixtures/bakerydemo.json
```diff
@@ -3531,6 +3852,54 @@
...
+      "approved_notifications": true,
+      "rejected_notifications": true,
+      "updated_comments_notifications": true,
+      "preferred_language": "ar",
+      "current_time_zone": "Asia/Beirut",
+      "avatar": "",
+      "dismissibles": {},
+      "theme": "system"
+    }
+  },
```
Stormheg, line 3902: While working on some extra demo data I noticed the user profile for the moderator user is missing here. The inactive user profile is missing too.

Is this intentional?

thibaudcolas: Hmm, `inactive` most likely yes but moderator no. I’ll fix this and merge.

bakerydemo/base/fixtures/bakerydemo.json
thibaudcolas: 👍 I’d say the more realistic the better. Real-world sites don’t have users named "editor".

thibaudcolas:
Thank you @Stormheg @laymonage! I switched the users’ documentation to a table so it’s simpler to compare the available options at a glance. Our new users are:

- Eddy Thorsørensen
- Jordan Rules
- Iman Absent
- Friedrich Wilhelm von Braunenberg
- Nour al-Hudā

I tried to add examples that were somewhat representative for German from Germany and Arabic from Lebanon speakers, with a mixture of lengths, and diacritics / non-latin characters.

laymonage:
OMG, Eddy Thor(sørensen) is now *canon*!

PR #439, opened 2023-08-14
Upgrade to Wagtail 5.1

Scotchester:
This is a barebones update that gets Bakery demo running Wagtail 5.1.x.

It does not (at least yet) do a few things we might want to consider doing:
- Switch some models from ModelAdmin to Snippets
  - Currently just makes the switch to the new wagtail_modeladmin pacakage
- Switch images to AVIF
  - Wasn't sure how valuable it would be to offer an example of using AVIF to people at this time, given that we'd still need to provide fallbacks for Edge
- Update the elasticsearch library to 7.x in requirements/production.txt
  - I wasn't entirely sure how to test doing this update


thibaudcolas (approved):
Thank you @Scotchester, this is all working great.

Issue #441, opened 2023-08-15
Replace ModelAdmin usage with snippets


thibaudcolas:
#439 moves the site to the wagtail-modeladmin package. We can go one step further and move to Snippets.

The site’s ModelAdmin usage is all in the one `base/wagtail_hooks.py`. Excerpt:

```python
class BreadIngredientAdmin(ModelAdmin):
    # These stub classes allow us to put various models into the custom "Wagtail Bakery" menu item
    # rather than under the default Snippets section.
    model = BreadIngredient
    search_fields = ("name",)
    inspect_view_enabled = True


class BreadTypeAdmin(ModelAdmin):
    model = BreadType
    search_fields = ("title",)


class BreadCountryAdmin(ModelAdmin):
    model = Country
    search_fields = ("title",)


class BreadModelAdminGroup(ModelAdminGroup):
    menu_label = "Bread Categories"
    menu_icon = "suitcase"  # change as required
    menu_order = 200  # will put in 3rd place (000 being 1st, 100 2nd)
    items = (BreadIngredientAdmin, BreadTypeAdmin, BreadCountryAdmin)

modeladmin_register(BreadModelAdminGroup)
```

Our official migration documentation: https://docs.wagtail.org/en/stable/reference/contrib/modeladmin/migrating_to_snippets.html

PR #444, opened 2023-08-24
Move all remaining ModelAdmin usage to Snippets

Stormheg:
Addresses https://github.com/wagtail/bakerydemo/issues/441

The models were already registered as snippets through the `@register_snippet` decorator on their definition.

I've replicated the ModelAdmin grouping in Snippets by creating custom SnippetViewSet and SnippetViewSetGroup classes. While at it, I've touched up the comments in the code a bit.

### DIFF
--- a/bakerydemo/breads/wagtail_hooks.py
+++ b/bakerydemo/breads/wagtail_hooks.py
@@ -1,0 +1,46 @@
+from wagtail.snippets.models import register_snippet
+from wagtail.snippets.views.snippets import SnippetViewSet, SnippetViewSetGroup
+
+from bakerydemo.breads.models import BreadIngredient, BreadType, Country
+
+
+class BreadIngredientSnippetViewSet(SnippetViewSet):
+    model = BreadIngredient
+    ordering = ("name",)
+    search_fields = ("name",)
+    inspect_view_enabled = True
+
+
+class BreadTypeSnippetViewSet(SnippetViewSet):
+    model = BreadType
+    ordering = ("title",)
+    search_fields = ("title",)
+
+
+class CountrySnippetViewSet(SnippetViewSet):
+    model = Country
+    ordering = ("title",)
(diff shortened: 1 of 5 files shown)


bakerydemo/breads/wagtail_hooks.py
```diff
@@ -0,0 +1,45 @@
+from wagtail.snippets.models import register_snippet
+from wagtail.snippets.views.snippets import SnippetViewSet, SnippetViewSetGroup
+
+from bakerydemo.breads.models import BreadIngredient, BreadType, Country
+
+
+class BreadIngredientSnippetViewSet(SnippetViewSet):
+    model = BreadIngredient
+    ordering = ("name",)
+    search_fields = ("name",)
```
laymonage, line 10: This is missing inspect view from the original `ModelAdmin` config

```suggestion
class BreadIngredientSnippetViewSet(SnippetViewSet):
    model = BreadIngredient
    ordering = ("name",)
    search_fields = ("name",)
    inspect_view_enabled = True
```

laymonage (approved):
I intentionally left these in #413 so that we still have a way to test ModelAdmin, but now that ModelAdmin is a separate package with its own tests included, I think it's OK to drop it from the bakerydemo.

Only one minor comment, but I can fix it as I merge. Nice to see that you moved these to the `breads` app, I've always wanted to do that. Thanks!

laymonage:
I changed the commits so that the first one does the replacing of `ModelAdmin` with `SnippetViewSet`, while the second one moves it to the `breads` app. I think that makes it easier to follow, see the diff in 89ebeb1af499fd22de2d8bfaa7946599d196d2fe for example.

bakerydemo/breads/wagtail_hooks.py
Stormheg: Thanks!

PR #445, opened 2023-08-24
Use MultipleChooserPanel instead of InlinePanel

Stormheg:
It allows selecting multiple authors at once and functions as a great demo of this panel.


laymonage (approved):
Looks good to me!

As a note, we can still test `InlinePanel`s in the `LocationPage` and `FormPage` pages.

PR #449, opened 2023-09-26
Add support for external links in search promotions.

TopDevPros:
Updated bakerydemo/templates/search/search_results.html to handle external links.


thibaudcolas (approved):
Looking great, thank you @TopDevPros

PR #452, opened 2023-10-02
Jinja version of bakerydemo

thibaudcolas:
Targets a *separate Jinja branch*, which will remain fully separate from `main`. I thought I would open this as a pull request for feedback and also so there’s a good record of exactly what is needed to get a site like `bakerydemo` converted to Jinja. @laymonage and I discussed keeping the separate `jinjademo` branch for future compatibility testing with Jinja, though we’d need further consensus from other maintainers on whether this is worthwhile or not.

This does not include an initial commit moving all templates from `templates/` to `jinja2/`.

Implementation notes:

- I chose to not use [django-jinja](https://niwi.nz/django-jinja/latest/) to avoid adding dependencies unless strictly necessary.
- I went for the "least changes possible" version of the Jinja implementation to make it easier to sync this Jinja version with `main` in the future


bakerydemo/base/jinja2tags.py
```diff
@@ -0,0 +1,61 @@
...
+    urlencode,
+)
+from django.contrib.staticfiles.storage import staticfiles_storage
+from django.urls import reverse
+from jinja2 import pass_context
+from jinja2.ext import Extension
+
+from wagtail.contrib.search_promotions.templatetags.wagtailsearchpromotions_tags import (
+    get_search_promotions,
+)
```
thibaudcolas, line 17: This is the only bakerydemo feature for which we don’t have Jinja support directly in Wagtail.

bakerydemo/blog/models.py
```diff
@@ -188,7 +188,7 @@ def tag_archive(self, request, tag=None):
             return redirect(self.url)
 
         posts = self.get_posts(tag=tag)
-        context = {"tag": tag, "posts": posts}
+        context = {"page": self, "tag": tag, "posts": posts}
```
thibaudcolas, line 191: Because of this not being set, the main bakerydemo doesn’t have breadcrumbs on blog tag indexes and doesn’t have the blog index as "active" in menus, which is likely unintended.

bakerydemo/jinja2/base.html
```diff
@@ -1,43 +1,37 @@
...
-                {% endif %}
+                {{ page.seo_title if page.seo_title else page.title }}
             {% endblock %}
             {% block title_suffix %}
-                | {{ settings.base.SiteSettings.title_suffix }}
+                | {{ settings("base.SiteSettings").title_suffix }}
             {% endblock %}
         </title>
-        <meta name="description" content="{% if page.search_description %}{{ page.search_description }}{% endif %}">
+        <meta name="description" content="{{ page.search_description if page and page.search_description }}">
```
thibaudcolas, line 13: This isn’t good in the base bakerydemo as views that aren’t page will have the meta tag with an empty value.

bakerydemo/jinja2/base/include/header-blog.html
```diff
@@ -19,7 +17,7 @@ <h1 class="index-header__title index-header__title--blog">{{ page.title }}</h1>
             {% endif %}
             {% if page.date_published %}
                 <div class="blog__published">
-                    {{ page.date_published }}
+                    {{ page.date_published|date("DATE_FORMAT") }}
```
thibaudcolas, line 20: I’m surprised Jinja doesn’t do this by default, perhaps there’s a way with configuration.

bakerydemo/jinja2/base/preview/person.html
```diff
@@ -1,9 +1,10 @@
 {% extends "base.html" %}
-{% load wagtailimages_tags %}
+
+{% block title %}{{ object.first_name }} {{ object.last_name }}{% endblock %}
```
thibaudcolas, line 3: Without this the page fails to render.

bakerydemo/jinja2/tags/gallery.html
```diff
@@ -1,13 +1,11 @@
...
     <div class="picture-card">
         <figure class="picture-card__image">
-            {% image img fill-645x480-c100 loading="lazy" %}
+            {{ image(img, "fill-645x480-c100", loading="lazy") }}
             <div class="picture-card__contents">
                 <p class="picture-card__title">{{ img.title }}</p>
             </div>
         </figure>
     </div>
-    {{ image.title }}
```
thibaudcolas, line 12: This looks odd and doesn’t seem to do anything even in the base demo site.

bakerydemo/jinja2/tags/top_menu.html
```diff
@@ -1,14 +1,12 @@
-{% load navigation_tags wagtailcore_tags %}
-{% get_site_root as site_root %}
-
+{% set menuitems=top_menu(parent=get_site_root(), calling_page=page).menuitems %}
 {% for menuitem in menuitems %}
-    <li class="presentation {{ menuitem.title|lower|cut:" " }}{% if menuitem.active %} active{% endif %}{% if menuitem.show_dropdown %} has-submenu{% endif %}">
+    <li class="presentation {{ menuitem.title|lower|cut(" ") }}{{ "active" if menuitem.active }} {{ "has-submenu" if menuitem.show_dropdown }}">
```
thibaudcolas, line 3: This "active" class does nothing currently so I’m not entirely sure why we have it.

bakerydemo/jinja2/base/form_page.html
```diff
@@ -12,22 +11,20 @@ <h1 class="index-header__title">{{ page.title }}</h1>
                 {% if page.intro %}
                     <p class="intro">{{ page.intro|richtext }}</p>
                 {% endif %}
-                {% if page.body %}
-                    {{ page.body }}
-                {% endif %}
+                {{ page.body if page.body }}
```
thibaudcolas, line 14: Hmm, this felt like a clever shortcut while writing this but maybe it’s better to stick with the previous syntax if we were to keep this Jinja version in sync.

bakerydemo/jinja2/base.html
```diff
@@ -1,43 +1,37 @@
...
-        <link rel="stylesheet" href="{% static 'css/font-marcellus.css' %}">
-        <link rel="stylesheet" href="{% static 'css/main.css' %}">
+        <link rel="stylesheet" href="{{ static('css/bootstrap.min.css') }}">
+        <link rel="stylesheet" href="{{ static('css/font-marcellus.css') }}">
+        <link rel="stylesheet" href="{{ static('css/main.css') }}">
     </head>
 
-    <body class="{% block body_class %}template-{{ self.get_verbose_name|slugify }}{% endblock %}">
-        {% wagtailuserbar %}
+    <body class="{% block body_class %}template-{{ page.get_verbose_name()|slugify if page }}{% endblock %}">
```
thibaudcolas, line 26: This `if` shouldn’t be needed once rebasing on #453.

thibaudcolas:
@laymonage this should _actually_ be ready to review now (but this is very low-priority).

thibaudcolas:
@laymonage ping

thibaudcolas:
If anyone needs this it’s available here: https://github.com/thibaudcolas/bakerydemo-jinja

PR #453, opened 2023-10-02
Templates cleanup

thibaudcolas:
Four bits of templates cleanup all spotted as part of #452. All of those changes are independent, but I chose to do it all as one PR for ease of review.

- [Remove unused templates](https://github.com/wagtail/bakerydemo/commit/9e6d196e11cd2e1833578bda9cc60c3ea3f6ea04) – as far as I could see there is no further code to remove for those.
- [Remove template output that does nothing](https://github.com/wagtail/bakerydemo/commit/d96b71372c71e87273fb581b3c076fb30cb0bf69) – outputs nothing when failing silently, breaks when _not_ failing silently.
- [Fix breadcrumbs missing on blog tag index](https://github.com/wagtail/bakerydemo/commit/5573fae2d83e8ee812be5b41edb6e627fa6673c6) – just a matter of setting the right context data.
- [Consistently override all page data in base template across all non-page views](https://github.com/wagtail/bakerydemo/commit/c022142dd34e6a240b170da36fe1f0d3e4993a90) – the more involved / opinionated change. Some aspects of this feel a bit overkill but I thought this would all be simpler to reason about if we just implemented all four `base` blocks everywhere (so `title`, `search_description`, `body_class`, `content`).

### DIFF
--- a/bakerydemo/templates/404.html
+++ b/bakerydemo/templates/404.html
@@ -2,6 +2,8 @@
 
 {% block title %}404 - Page not found{% endblock %}
 
+{% block search_description %}Sorry, this page could not be found{% endblock %}
+
 {% block body_class %}template-404{% endblock %}
 
 {% block content %}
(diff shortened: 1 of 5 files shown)


lb- (approved):
Looks good to me, great finds.

PR #454, opened 2023-10-09
Remove unnecessary variables from base.html

rypptc:
I was revising the code in base.html and I think the variables in line 35 are not needed for the navigation menu to work, since they are already defined in header.html. I tested the site deleting the variables and the menu renders fine.


thibaudcolas (approved):
Thank you @rypptc! Just tested and this does work fine without them.

Issue #421, opened 2023-06-26
Replace flake8 and isort with ruff


laymonage:
Per wagtail/wagtail#10324.

Need to consider wagtail/wagtail#10551 and wagtail/wagtail#10556 as well.

thibaudcolas:
Done in #461

PR #461, opened 2023-11-10
Updated project requirements

jsma:
Some of these were necessary for proper Django 4.2 support. Replaced linters with `ruff` and updated the `pre-commit` configuration to match upstream Wagtail for consistency.

`bakerydemo` does not appear to have its own test suite, but so far it appears to run correctly ;)


lb-:
Just fired off the CI - it's failing so may need some tweaks. 

Thanks for this @jsma - it will be nice to use the same linting tooling as Wagtail.

jsma:
Apologies for missing that, I brought the Makefile in line with upstream for the relevant sections. 🤞it passes this time. Thanks!

jsma:
This looks to be working now, and should address #421

.pre-commit-config.yaml
```diff
@@ -7,45 +7,45 @@ repos:
...
+          - '@typescript-eslint/eslint-plugin@6.2.1'
+          - '@typescript-eslint/parser@6.2.1'
+          - '@wagtail/eslint-config-wagtail@0.4.0'
   - repo: https://github.com/thibaudcolas/pre-commit-stylelint
-    rev: v14.2.0
+    rev: v15.10.3
     hooks:
       - id: stylelint
-        files: \.css$
+        files: \.scss$
```
lb-, line 43: I think we only use CSS in this project.

jsma: No copy/paste job shall go unpunished ;)

jsma:
Let me know your thoughts but perhaps #463 should be merged first and then I can rebase off of that to straighten out pre-commit-config.yaml. I also have [WIP](https://github.com/GreenLightGo/bakerydemo/tree/ruff-format) to introduce `ruff format` to match https://github.com/wagtail/wagtail/pull/11220 This PR intentionally kept `ruff` at the version currently used by Wagtail until my upstream PR is approved. If the upstream PR gets merged before this one, I could lump the ruff upgrade into this branch or just submit a follow up PR.

lb-:
Good call @jsma  - I did not realise there would be a large overlap with #463 but if this one does get merged first, I am happy to rebase.

.pre-commit-config.yaml
```diff
@@ -7,45 +7,45 @@ repos:
...
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.5.1
+    rev: v3.0.3
     hooks:
       - id: prettier
-        types_or: [css, javascript, json, yaml]
+        types_or: [css, scss, javascript, ts, tsx, json, yaml]
+        additional_dependencies:
+          # Keep in sync with package.json
+          - prettier@3.0.1
```
thibaudcolas, line 25:
```suggestion
        types_or: [css, javascript, json, yaml]
```

I don’t think we want to sync this with Wagtail to _that_ extent. First off `additional_dependencies` is useless if we install Prettier only, and for this project we wouldn’t currently use SCSS / TypeScript.

.pre-commit-config.yaml
```diff
@@ -7,45 +7,45 @@ repos:
...
-        files: \.(js)$
+        files: \.(js|ts|tsx)$
         args: [--report-unused-disable-directives]
         additional_dependencies:
-          - eslint@8.8.0
+          # Keep in sync with package.json
+          - eslint@8.46
+          - '@typescript-eslint/eslint-plugin@6.2.1'
+          - '@typescript-eslint/parser@6.2.1'
+          - '@wagtail/eslint-config-wagtail@0.4.0'
```
thibaudcolas, line 38: Same as above. We don’t need such complex JS linting for this project.

.pre-commit-config.yaml
```diff
@@ -7,45 +7,45 @@ repos:
...
     hooks:
       - id: stylelint
         files: \.css$
         additional_dependencies:
-          - stylelint@14.9.1
-          - stylelint-config-standard@26.0.0
-          - stylelint-config-prettier@9.0.3
+          # Keep in sync with package.json
+          - stylelint@14.2.0
+          - '@wagtail/stylelint-config-wagtail@0.5.0'
```
thibaudcolas, line 47: Same as above.

thibaudcolas (approved):
Looking solid, thank you @jsma! I see our setup on Wagtail has changed since (more ruff), but this is a step in the right direction so let’s merge.

thibaudcolas:
Merged locally as I’m not allowed to push to this fork

.pre-commit-config.yaml
jsma: I was going too fast and generally being a copy/paste fool. I did mean to circle back and add `ruff format` and upgrade ruff now that Wagtail is doing the same. Another PR for another day.

PR #473, opened 2024-01-02
Upgrade to Django 5.0 + add FormSubmissionsPanel

thibaudcolas:
I believe the only change strictly needed is the one to the requirements’ definition, but I took the time to also update Django settings for v5.0 based on what was deprecated / removed. And updating all the documentation URLs so they’re more "evergreen".

I also added `FormSubmissionsPanel` to our `FormPage` (see https://github.com/wagtail/wagtail/issues/11405).


Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN set -ex \
         zlib1g-dev \
     " \
     && apt-get update && apt-get install -y --no-install-recommends $BUILD_DEPS \
-    && python3.9 -m venv ${VIRTUAL_ENV} \
+    && python3.12 -m venv ${VIRTUAL_ENV} \
     && pip install -U pip \
     && pip install --no-cache-dir -r /requirements/production.txt \
```
zerolab, line 38: nitpick: would be good to change these to be `python3.12 -m pip`

laymonage (commented):
Some of the changes in this PR have been picked up in others (#527, #494, 4fa8d9ae8d0c8bb188ba97ac321a4db604f93a9a). I'll rebase and have another look.

bakerydemo/settings/base.py
```diff
@@ -149,21 +149,27 @@
...
 
 LANGUAGE_CODE = "en-us"
 
 TIME_ZONE = "UTC"
 
 USE_I18N = True
 
-USE_L10N = True
-
-USE_TZ = True
```
laymonage, line 162: Even though `USE_TZ` defaults to `True` in Django 5.0+, it's still included in the default template: https://github.com/django/django/blob/240421c7c4c81fe5df26274b807266bd4ca73d7f/django/conf/project_template/project_name/settings.py-tpl#L111

So I think we can keep this in.

laymonage (approved):
LGTM, thanks for updating the docs links to use `stable`!

PR #474, opened 2024-01-05
Update header.html

Ian-Jhon:
set the landmark of navigation_logo to nav;
set the value of navigation_logo to `{{ settings.base.SiteSettings.title_suffix }}`.


lb- (changes requested):
Could you maybe review the code, if looks like no closing tags have been added. 

Also, I'm not sure this approach is right. The title suffix is for the html title (as in the browser tab title), not the navigation title.

gasman:
Closing as invalid

PR #475, opened 2024-01-26
Add list_export for Snippets

thibaudcolas:
Available in v5.2 so this will work fine with current and future Wagtail


laymonage (approved):
Thanks!

PR #478, opened 2024-02-07
Add verbose name to Country model

lb-:
- Supports testing for edge case issues such as [#10982](https://github.com/wagtail/wagtail/issues/10982) & listing labels


lb-:
Screenshots where this is used for reference

Screenshot 2024-02-08 at 8 11 58 am
Screenshot 2024-02-08 at 8 11 23 am

bakerydemo/breads/models.py
```diff
@@ -28,6 +28,7 @@ def __str__(self):
         return self.title
 
     class Meta:
+        verbose_name = "Country of Origin"
         verbose_name_plural = "Countries of Origin"
```
laymonage, line 32: Something that's been bugging me for a while in this project, the verbose names shouldn't be title-cased. We should let Django/Wagtail decide when it needs to be capitalised (e.g. using `capfirst()`), as per [Django's recommendation](https://docs.djangoproject.com/en/5.0/topics/db/models/#:~:text=If%20the%20verbose%20name%20isn,name%2C%20converting%20underscores%20to%20spaces.&text=The%20convention%20is%20not%20to,letter%20where%20it%20needs%20to.)

Would be nice to have one migration that changes this for all models, or maybe even just edit the old migration files as this doesn't really change anything in the database.

lb-: OK, should we close this PR and I can raise an issue for that global change instead?

PR #479, opened 2024-02-07
Black format settings/production

lb-:
When running `make format-server` this file gets updated, I thought it would be good for this to be in the main branch.


lb-:
OK not sure what's going on, maybe a Black version difference. Ignore this PR.

PR #481, opened 2024-03-17
add gallery links

aurelpere:


lb-:
Could you please format the HTML, the nesting is a bit off. `make format-server` should do it or just manually adjust the nesting.
Thanks @aurelpere

Issue #390, opened 2023-02-02
Instruction in readme.md not clear


Orange-Tofu:
### Problem
In the readme.md file, under the subheading [Setup with Virtualenv](https://github.com/wagtail/bakerydemo/#setup-with-virtualenv).

The 4th set of commands does not run directly in windows command prompt. i.e. the command:
`$ cp bakerydemo/settings/local.py.example bakerydemo/settings/local.py`.
People not familiar with linux will not know `cp` is a linux command.

### Possible solution
`cp` can either run on windows powershell or using the cygwin package.

PR #482, opened 2024-03-22
Update documentations -- Issue #390

TimothyElems:
[Instruction in readme.md not clear #390](https://github.com/wagtail/bakerydemo/issues/390)

Updated and made changes to <Contributions.md>, to make it compliant with newer Markdown syntax

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -5,7 +5,7 @@ agree to abide by the Wagtail [code of conduct](https://github.com/wagtail/wagta
 
 If you do not have write access to the repo, fork, then clone the repo:
 
-    git clone git@github.com:your-username/bakerydemo.git
+    `$ git clone git@github.com:your-username/bakerydemo.git`
 
 Set up your machine for [development](https://github.com/wagtail/bakerydemo/blob/master/readme.md).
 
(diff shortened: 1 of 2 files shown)


contributing.md
```diff
@@ -5,7 +5,7 @@ agree to abide by the Wagtail [code of conduct](https://github.com/wagtail/wagta
 
 If you do not have write access to the repo, fork, then clone the repo:
 
-    git clone git@github.com:your-username/bakerydemo.git
+    `$ git clone git@github.com:your-username/bakerydemo.git`
```
gasman, line 8: This change doesn't seem to be correct - this is already a code block, so the backtick characters appear in the output. Also, we should leave out the `$` - it should already be clear that this is a command to enter at the command prompt, and inexperienced users might think that the `$` is part of the command.

readme.md
```diff
@@ -133,35 +133,41 @@ You can run the Wagtail demo locally without setting up Vagrant or Docker and si
 
 With [PIP](https://github.com/pypa/pip) and [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/)
 installed, run:
-
+```bash
     mkvirtualenv wagtailbakerydemo
```
gasman, line 137: Please remove the indenting from these blocks - the backtick notation eliminates the need for it, so leaving it in place means that they're indented too much.

gasman (changes requested):
Thanks @TimothyElems!

contributing.md
TimothyElems: Thanks for the feedback! I've made some changes!

readme.md
TimothyElems: Fixed this!

PR #485, opened 2024-04-23
Add more configuration for testing various features

laymonage:
- Add missing fixtures for German and Arabic locales
- Add more filters to snippet viewsets
- Register Country model with a `ModelViewSet` instead of a `SnippetViewSet`, so we have an example of a `ModelViewSet`
  - Enable inspect view for `CountryModelViewSet`
- Add `UserApprovalTask`, so we have an example of different task types
  - Based on https://docs.wagtail.org/en/stable/extending/custom_tasks.html
- Fix verbose name of user-facing models
  - So we can make sure we're applying `capfirst()` in the right places in Wagtail core, and prevent unnecessary capitalisation in other places. See https://github.com/wagtail/bakerydemo/pull/478#discussion_r1482662326.


bakerydemo/base/migrations/0020_alter_footertext_options.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 5.1.dev20240316121317 on 2024-04-23 04:11
```
thibaudcolas, line 1: 🤔 living on the bleeding edge

bakerydemo/base/models.py
```diff
@@ -510,3 +514,58 @@ class SiteSettings(BaseSiteSetting):
     panels = [
         FieldPanel("title_suffix"),
     ]
+
+
+class UserApprovalTaskState(TaskState):
+    pass
+
+
+class UserApprovalTask(Task):
```
thibaudcolas, line 523:
```suggestion
class UserApprovalTask(Task):
    """
    Based on https://docs.wagtail.org/en/stable/extending/custom_tasks.html.
    """
    
```

PR #486, opened 2024-05-01
Bump to Wagtail 6.1

Stormheg:
Looks like bakerydemo is not affected by any of the upgrade considerations: https://docs.wagtail.org/en/stable/releases/6.1.html


lb- (approved):
Thanks @Stormheg

PR #488, opened 2024-06-12
Fix explanation of .env and local.py

gasman:
As of d7a607a26e5190bb49048f7041eef63ecea8f1e4 both local.py and .env are initially empty, as the bakerydemo.settings.dev module is directly usable as DJANGO_SETTINGS_MODULE. Update the explanation in readme.md to reflect this.

(Maybe this is too much waffly background detail to belong inside a set of step-by-step instructions? However, I think there's some value in mentioning that we use `django-dotenv`, since we sometimes see people freestyling the setup and installing the similar-but-different `python-dotenv`, then wondering why it doesn't work...)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -154,8 +154,7 @@ cd bakerydemo
 pip install -r requirements/development.txt
 ```
 
-Next, we'll set up our local environment variables. We use [django-dotenv](https://github.com/jpadilla/django-dotenv)
-to help with this. It reads environment variables located in a file name `.env` in the top level directory of the project. The only variable we need to start is `DJANGO_SETTINGS_MODULE`:
+Next, we need to create the files `.env` and `bakerydemo/settings/local.py`, which provide a place for local configuration settings that need to be kept outside of version control. No such settings are required for a standard installation, but warnings will be displayed if these files are not present:
 ```bash
 cp bakerydemo/settings/local.py.example bakerydemo/settings/local.py
 cp .env.example .env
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -154,8 +154,7 @@ cd bakerydemo
 pip install -r requirements/development.txt
 ```
 
-Next, we'll set up our local environment variables. We use [django-dotenv](https://github.com/jpadilla/django-dotenv)
-to help with this. It reads environment variables located in a file name `.env` in the top level directory of the project. The only variable we need to start is `DJANGO_SETTINGS_MODULE`:
+Next, we need to create the files `.env` and `bakerydemo/settings/local.py`. The `.env` file is used by [django-dotenv](https://github.com/jpadilla/django-dotenv) to store environment variables relevant to the local installation, such as `DJANGO_SETTINGS_MODULE`. Similarly, `bakerydemo/settings/local.py` can be used to store local Django settings such as database connection details that need to be kept outside of version control. Both of these files are initially empty, as the bakerydemo project provides usable defaults for all settings, but they must exist for the server to start:
```
laymonage, line 157: > but they must exist for the server to start

I think the server would still start without `.env`, but `django-dotenv` will print a warning in the console. We use `try`/`except` when importing `local.py` (e831dcee2790c22c597745b89b9bb28ec2edaa63), so it's not required too, but I still think it's worthwhile to keep the instructions for adding these files so they don't have to figure this out later.

Maybe "but they must exist for the server to start gracefully" would be more appropriate?

gasman:
@laymonage Have now rewritten this to be more accurate about the missing file only producing a warning, and to move the "how you would actually make use of this" details to a separate notes section so that it doesn't derail the setup instructions.

laymonage (approved):
Looks good to me, thanks!

Issue #493, opened 2024-07-31
Fix all remaining design and coding accessibility issues


thibaudcolas:
Follow-up to [wagtail#12157](https://github.com/wagtail/wagtail/issues/12157). We want the bakerydemo website to be free of design and coding related accessibility issues. Content issues are desirable so we can test the accessibility checker.

- [x] Fix color contrast / link not distinguishable in text issues showing up on recipe pages
- [x] Fix main navigation menu issues when a top-level menu item has sub pages configured to show in the menu (create a page as a child of "Contact us"). Or remove support for dropdown menus
- [x] (test all other pages of the site and report possible improvements here / fix the issues)

albinazs:
**Fixed issues:**

[x] Fix color contrast and link not distinguishable in text issues showing up on Recipe pages
[x] Remove support for dropdown menus within the main navigation menu
[x] Fix the aria-allowed attributes issue on the Contact us form page
[x] Fix the missing table headers issue on the Breads index page
[x] Fix the missing heading level 1 issue on the Recipes index page
[x] Add missing headings to the Location pages
[x] Update Twitter to X on the FE only (icon, label, field verbose name for admin)

**Other possible improvements:** 

[] Add a skip main navigation link
[] Fix not visible focus appearance for the `.picture-card` (used for blog items on the home page, location items on the location index page)

albinazs:
**Additional fixed issue:**

[x] Add a skip main navigation link

PR #503, opened 2024-08-20
Fix small accessibility and design issues

albinazs:
Addresses #493 

- [x] Fix color contrast and  link not distinguishable in text issues showing up on Recipe pages
- [x] Remove support for dropdown menus within the main navigation menu 
- [x] Fix the aria-allowed attributes issue on the Contact us form page
- [x] Fix the missing table headers issue on the Breads index page
- [x] Fix the missing heading level 1 issue on the Recipes index page
- [x] Add missing headings to the Location pages
- [x] Update Twitter to X on the FE only (icon, label, field verbose name for admin)

### DIFF
--- a/bakerydemo/templates/tags/top_menu.html
+++ b/bakerydemo/templates/tags/top_menu.html
@@ -2,13 +2,7 @@
 {% get_site_root as site_root %}
 
 {% for menuitem in menuitems %}
-    <li class="presentation {{ menuitem.title|lower|cut:" " }}{% if menuitem.active %} active{% endif %}{% if menuitem.show_dropdown %} has-submenu{% endif %}">
-        {% if menuitem.show_dropdown %}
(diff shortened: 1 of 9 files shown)


bakerydemo/templates/tags/top_menu.html
```diff
@@ -2,13 +2,7 @@
 {% get_site_root as site_root %}
 
 {% for menuitem in menuitems %}
-    <li class="presentation {{ menuitem.title|lower|cut:" " }}{% if menuitem.active %} active{% endif %}{% if menuitem.show_dropdown %} has-submenu{% endif %}">
-        {% if menuitem.show_dropdown %}
-            <a href="{% pageurl menuitem %}" class="allow-toggle">{{ menuitem.title }} <span><a class="caret-custom dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false"></a></span></a>
```
thibaudcolas, line 7: It looks like there are classes here that could also be removed from the CSS (`dropdown-menu`, `has-submenu`, `caret-custom`)

thibaudcolas (approved):
Thank you, all looking great!

PR #505, opened 2024-08-28
link to virtualenv corrected

wijnandb:
It linked to https://virtualenv.pypa.io/en/stable/installation/ which gave a 404, link has been changed to https://virtualenv.pypa.io/en/stable/installation.html


gasman:
Thanks @wijnandb!

PR #508, opened 2024-10-05
Adding venv installation documentation

ViragJain3010:
Since Python includes the `venv` module by default, there's no need for the `virtualenv` package, which adds unnecessary overhead. Furthermore, Wagtail’s official documentation also recommends using venv, so it's advisable to follow this approach.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -33,11 +33,11 @@ This demo is aimed primarily at developers wanting to learn more about the inter
 - [Gitpod](#setup-with-gitpod)
 - [Vagrant](#setup-with-vagrant)
 - [Docker](#setup-with-docker)
-- [Virtualenv](#setup-with-virtualenv)
+- [Venv](#setup-with-venv)
 
 If you want to see what Wagtail is all about, we suggest trying it out through [Gitpod](#setup-with-gitpod).
 If you want to set up Wagtail locally instead, and you're new to Python and/or Django, we suggest you run this project on a Virtual Machine using [Vagrant](#setup-with-vagrant) or [Docker](#setup-with-docker) (whichever you're most comfortable with). Both Vagrant and Docker will help resolve common software dependency issues.
-Developers more familiar with virtualenv and traditional Django app setup instructions should skip to [Setup with virtualenv](#setup-with-virtualenv).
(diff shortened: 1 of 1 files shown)


ViragJain3010:
@laymonage @gasman

Scotchester:
@ViragJain3010 Please don't tag specific people unless you've already been in discussion with them and they're expecting it.

readme.md
```diff
@@ -119,18 +119,38 @@ To tail the logs from the Docker containers in realtime, run:
...
+```bash
+wagtailbakerydemo\Scripts\activate.bat
+# if wagtailbakerydemo\Scripts\activate.bat doesn't work, run:
+wagtailbakerydemo\Scripts\Activate.ps1
+
+```
+Note: On Ubuntu systems, the [Venv](https://docs.python.org/3/library/venv.html) package may not be installed by default. To install it, use the following command:
+```bash
+sudo apt install python3-venv
+```
```
thibaudcolas, line 150:
```suggestion
```

I see this is absent from the Django docs, which we reference as our reason for using this setup. And our docs above mention "Developers more familiar with venv" when directing people to this section. So I think we can skip this.

readme.md
```diff
@@ -119,18 +119,38 @@ To tail the logs from the Docker containers in realtime, run:
...
+wagtailbakerydemo\Scripts\Activate.ps1
+
+```
+Note: On Ubuntu systems, the [Venv](https://docs.python.org/3/library/venv.html) package may not be installed by default. To install it, use the following command:
+```bash
+sudo apt install python3-venv
+``` 
+
+### Alternatively
+
```
thibaudcolas, line 153: I don’t think retaining this is needed. As you mention the Wagtail docs only cover `venv`. We can trust that users of `virtualenvwrapper` would know how to do those steps below.

readme.md
```diff
@@ -146,6 +166,8 @@ mkvirtualenv wagtailbakerydemo --python=python3.12
 python --version
 ```
 
+# Essential Setup
+
```
thibaudcolas, line 170:
```suggestion
```

This part of the setup is only relevant for virtual environment users so shouldn’t have its own section.

thibaudcolas (approved):
Nice!

PR #517, opened 2024-11-19
More human-friendly WAGTAIL_SITE_NAME

gasman:
As of Wagtail 6.3 `WAGTAIL_SITE_NAME` is used as the heading for the admin dashboard, so let's make it a human-friendly title (and avoid creating the temptation for developers to try and munge the capitalisation in code, as we did in [#12592](https://github.com/wagtail/wagtail/issues/12592) :-) )
Screenshot 2024-11-19 at 14 48 11

### DIFF
--- a/bakerydemo/settings/base.py
+++ b/bakerydemo/settings/base.py
@@ -194,7 +194,7 @@ WAGTAILSEARCH_BACKENDS = {
 }
 
 # Wagtail settings
-WAGTAIL_SITE_NAME = "bakerydemo"
+WAGTAIL_SITE_NAME = "The Wagtail Bakery"
 
 WAGTAIL_I18N_ENABLED = True
 


laymonage (approved):
It slightly irks me when we call this "(The) Wagtail Bakery" (without the "demo"), since we also have https://github.com/wagtail-nest/wagtail-bakery (which is older than the demo).

I know this is the name we use in the front-facing site's header though, so I suppose it's just my pet peeve 😂

gasman:
@laymonage IIRC, it was originally my suggestion to give it the developer-facing name "bakerydemo" to avoid that naming collision. I think it's fair game to call it "The Wagtail Bakery" in user-facing text within the demo itself though - think of it as roleplaying in a fictional universe where the Icelandic bakery exists but the package doesn't :-)

Issue #518, opened 2024-11-19
Remove `from __future__ import unicode_literals`


lb-:
I think we can now remove the `from __future__ import unicode_literals` declarations in a few files as we are now fully migrated to Python 3.*  (and no longer support older versions of Python in Wagtail).

https://github.com/search?q=repo%3Awagtail%2Fbakerydemo%20from%20__future__%20import%20unicode_literals&type=code

See 
- https://python-future.org/unicode_literals.html
- https://docs.python.org/3/library/__future__.html

input:
Hi @lb-
I'll do this as a first issue.

lb-:
Thanks @input - nice username by the way, how did you snag that?

input:
Thanks - signed up a long time ago 😄

PR #519, opened 2024-11-30
Remove 'from __future__ import unicode_literals' statements (issue #518)

input:
Removed all `from __future__ import unicode_literals` statements.


lb- (approved):
Thank you @input  - tested locally and it's working fine.

PR #521, opened 2025-01-23
Add block previews

laymonage:
Some blocks are intentionally left out, to test the behaviour of blocks that do not have previews enabled.

Different approaches are simulated:
- Passing `preview_value`, `description` via init kwargs
- Setting them in `Meta`
- Overriding `get_preview_value()`
- Custom `preview_template` to showcase using a static image for the preview

I didn't find a use case for overriding `get_preview_context()`, but let me know if we'd like to have it anyway.

https://github.com/user-attachments/assets/4fdc23c9-4f3c-4dde-b3f3-62371b98b46f

https://github.com/user-attachments/assets/761d2c07-ffed-4b38-94ba-eec45ffffe2b

### DIFF
--- a/bakerydemo/recipes/blocks.py
+++ b/bakerydemo/recipes/blocks.py
@@ -39,7 +39,19 @@ class RecipeStreamBlock(StreamBlock):
         icon="pilcrow", template="blocks/paragraph_block.html", group="Content"
     )
     block_quote = BlockQuote(group="Content")
-    table_block = TableBlock(group="Content")
+    table_block = TableBlock(
+        group="Content",
+        description="A table of data with plain text cells",
+        preview_value={
+            "first_row_is_table_header": "True",
+            "data": [
+                ["Bread type", "Origin"],
+                ["Anpan", "Japan"],
+                ["Crumpet", "United Kingdom"],
+                ["Roti buaya", "Indonesia"],
+            ],
+        },
+    )
     typed_table_block = TypedTableBlock(
         [
             ("text", CharBlock()),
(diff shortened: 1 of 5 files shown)


bakerydemo/recipes/blocks.py
```diff
@@ -29,6 +29,13 @@ class Meta:
         icon = "tick"
 
 
+class CustomTypedTableBlock(TypedTableBlock):
+    def get_preview_value(self):
+        # Avoid having to construct TypedTable ourselves. This won't be
+        # necessary if https://github.com/wagtail/wagtail/pull/12808 is merged
+        return self.to_python(self.meta.preview_value)
```
laymonage, line 36: Ref: https://github.com/wagtail/wagtail/pull/12808

bakerydemo/templates/base/preview/static_embed_block.html
```diff
@@ -0,0 +1,6 @@
+{% extends "wagtailcore/shared/block_preview.html" %}
+{% load static %}
+
+{% block content %}
+    <img class="static-block-preview" src="{% static "img/embed-block-preview.png" %}" alt="Preview of an embedded video" />
+{% endblock %}
```
laymonage, line 6: I thought it'd be nice to showcase how a static image preview could work with the current implementation.

thibaudcolas: I agree, but I’d rather we demo-ed with an SVG so it’s less file to load and will be easier to dark-theme-ify

thibaudcolas (approved):
A few things I’d do a bit differently but all working well! And this doesn’t interfere with Wagtail 6.3 so safe to merge already. I’ll do the tweaks I suggested and merge.

bakerydemo/recipes/blocks.py
laymonage: Yes, but it's currently still in `main`, so this is still necessary for stable 6.4.x! 😬

PR #522, opened 2025-01-26
Updated the old recommended installation approach for django with the…

ankit02327:
-The old link was stating "This document is for an insecure version of Django that is no longer supported. Please upgrade to a newer release!".
-Updated with the link to latest documentation.


gasman:
Thanks @ankit02327!

PR #525, opened 2025-01-31
Update boto3 and remove explicit botocore dependency

laymonage:
- https://docs.djangoproject.com/en/stable/internals/deprecation/#deprecation-removed-in-5-1 https://github.com/wagtail/bakerydemo/blob/5e261b5b5defc2297ff37cdb9519e856950dc843/requirements/base.txt#L1
- Update boto3 to fix https://github.com/boto/boto3/issues/3766
- Remove explicit botocore dependency, as boto3 already depends on it


laymonage (commented):
Updated to only bump the `boto3` and dropping `botocore` dependency, as the `STORAGES` setting update has been done in #527.

PR #527, opened 2025-02-19
Use Django 4.2+ syntax for storage

RealOrangeOne:
The previous settings are deprecated, meaning the bakerydemo will always use `FilesystemStorage`, which doesn't work in production.


bakerydemo/settings/base.py
```diff
@@ -182,6 +182,15 @@
 MEDIA_ROOT = os.path.join(PROJECT_DIR, "media")
 MEDIA_URL = "/media/"
 
+STORAGES = {
+    "default": {
+        "BACKEND": "django.core.files.storage.FileSystemStorage",
+    },
+    "staticfiles": {
+        "BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage",
```
laymonage, line 190: I think we should use `ManifestStaticFilesStorage` as the default, to match Wagtail's default project template

bakerydemo/base/management/commands/reset_demo.py
```diff
@@ -17,7 +17,7 @@ def handle(self, **options):
             )
 
         # 1. (optional) Remove all objects from S3
-        if "s3" in settings.DEFAULT_FILE_STORAGE:
```
laymonage, line 20: Good catch!

laymonage (commented):
One comment about using `ManifestStaticFilesStorage`, but otherwise this looks good to me! I'll update and merge. Thanks!

laymonage (approved):

PR #544, opened 2025-06-19
Add API fields and usage example

### LINKED ISSUE #83: Demonstrate API
Not sure we want to tackle this for v1, but we had discussed before the sprint the possibility of demonstrating the Wagtail REST API. 

We could consume an endpoint of breads with a simple jquery demo, or be more modern and use React, though that might involve bringing in webpack (though I tried a React demo once that provided a runtime compilation rather than using webpack - could look that up). 

Whatever we do, I feel we should keep it simple.

laymonage:
I forgot to generate migrations after #521, so the first commit adds them.

Fixes #83.


thibaudcolas (approved):
Nice!

PR #551, opened 2025-08-13
implement new django csp baseline

Chiemezuo:
## Introduction
From [Django 6.0](https://docs.djangoproject.com/en/dev/releases/6.0/) which should be released in December 2025, Django will have an in-built CSP package, and will no longer need the external [django-csp](https://django-csp.readthedocs.io/en/latest/) package.

This PR, although small, has some months to go before being ready for a merge, but it's good to have this under our radar.

### DIFF
--- a/bakerydemo/settings/base.py
+++ b/bakerydemo/settings/base.py
@@ -13,6 +13,7 @@ https://docs.djangoproject.com/en/stable/ref/settings/
 import os
 
 import dj_database_url
+from django.utils.csp import CSP
 
 # Build paths inside the project like this: os.path.join(PROJECT_DIR, ...)
 PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -1,4 +1,4 @@
-Django>=5.2,<5.3
+Django>=6.0,<7.0
```
thibaudcolas, line 1:
```suggestion
Django>=5.2,<6.0
```

I believe this will allow us to have working installs for most users, and for early adopters support installing release candidates of Django 6.0

bakerydemo/settings/base.py
```diff
@@ -226,30 +227,14 @@
...
-        CSP_REPORT_URI = os.environ.get("CSP_REPORT_URI")
+    MIDDLEWARE.append("django.middleware.csp.ContentSecurityPolicyMiddleware")
+    # Gravatar images should be disabled for strict CSP
+    WAGTAIL_GRAVATAR_PROVIDER_URL = None
+
+# To enforce a CSP policy in report-only mode:
+SECURE_CSP_REPORT_ONLY = {
+    "default-src": [CSP.SELF, "*.wagtail.org"],
+    # Add more directives as needed.
+}
```
thibaudcolas, line 238: Could you update this so we can preserve how this is done with environment variables? I think ideally we would want to preserve our ability to configure this with environment variables, just do this with the vanilla Django implementation.

I think this should also be done only when `if "CSP_DEFAULT_SRC" in os.environ:`.

bakerydemo/settings/base.py
```diff
@@ -13,6 +13,7 @@
 import os
 
 import dj_database_url
+from django.utils.csp import CSP
```
thibaudcolas, line 16: Is it possible for us to move this and the middleware addition within a Django version check, so we can merge this while still supporting Django 5.2?

bakerydemo/settings/base.py
```diff
@@ -226,30 +227,14 @@
...
-        CSP_BASE_URI = os.environ.get("CSP_BASE_URI").split(",")
-    if "CSP_OBJECT_SRC" in os.environ:
-        CSP_OBJECT_SRC = os.environ.get("CSP_OBJECT_SRC").split(",")
-    if "CSP_FRAME_SRC" in os.environ:
-        CSP_FRAME_SRC = os.environ.get("CSP_FRAME_SRC").split(",")
-    if "CSP_REPORT_URI" in os.environ:
-        CSP_REPORT_URI = os.environ.get("CSP_REPORT_URI")
+    MIDDLEWARE.append("django.middleware.csp.ContentSecurityPolicyMiddleware")
+    # Gravatar images should be disabled for strict CSP
+    WAGTAIL_GRAVATAR_PROVIDER_URL = None
```
thibaudcolas, line 232: I think we would want to keep using Gravatar, it’s just a matter of configuration no? We can allow loading images from there while still having a strict CSP.

thibaudcolas (changes requested):
Nice! This seems pretty mergeable to me if those changes are limited to the one settings and requirements files. I think it’d make sense to merge this as soon as Django 6.0 alpha 1 is released, possibly even before, so we can try this out on the `main` branch of bakerydemo. Just making sure any loading of Django 6.0 APIs are behind a Django version check.

Also we would need to remove the dependency on `django-csp`.

requirements/base.txt
Chiemezuo: With the `<6.0`, wouldn't that mean that it would not be possible to install a release candidate for `6.0`?

bakerydemo/settings/base.py
Chiemezuo: Yes, we can do this. But we would be trusting all the content from Gravatar to not be malicious. What do you think?

requirements/base.txt
Chiemezuo: Also, we could remove the dependency on `django-csp` when `6.0` is available. For now, I suppose we could leave it, though.

PR #552, opened 2025-08-15
Update Dockerfile to have Debian Trixie compatible packages

ArneTR:
The underlying default base image for python:3.12-slim was updated and some debian packages have changed


gasman:
Can confirm that this successfully fixes the Docker build. Thanks @ArneTR!

PR #553, opened 2025-08-20
Fix UI truncation for long image titles with ellipsis and tooltips

Sunil1532:
## Problem
Long image file names and titles can cause UI layout issues in image choosers, listings, and gallery displays. The current implementation doesn't handle text overflow gracefully.

## Solution
- Added `.truncate-text` CSS class with ellipsis handling
- Applied truncation to picture card titles
- Added tooltip functionality showing full title on hover
- Ensures clean UI layout regardless of title length

## Testing
- Uploaded images with very long file names
- Verified truncation displays properly with "..." 
- Confirmed full title appears in tooltip on hover
- No layout breaks observed

Addresses issue with challenging image file names display in choosers and listings.


Sunil1532:
Sorry for that . Actually its my first contribution  and i misunderstood the issue. I would to like to add ui enhancement for bakery demo project like adding dark theme and light theme toggle button for background and  adding features to remaining content section

Issue #545, opened 2025-06-20
Demonstrate template-level caching


thibaudcolas:
A common scenario, which our demo site somehow doesn’t demo at all! See [Template fragment caching](https://docs.wagtail.org/en/stable/topics/writing_templates.html#template-fragment-caching). This would involve setting up:

- `{% wagtailcache %}`
- And-or potentially `{% wagtailpagecache %}`

PR #561, opened 2025-10-20
Feature/545 demonstrate template level caching

lupyana:
- I have added examples on how to use wagtailcache functionality to the blog_index_page.html, where I have cached the blog tags and blog posts, doing this showed a n improvement in load times as shown in the attached screenshots bellow from the debug toolbar

Screenshot 2025-10-20 at 14 30 21

Screenshot 2025-10-20 at 14 47 21

- Also added wagtailpagecache example in the blog_page.html, this also showed a significant improvement in the load times by utilizing the pagecache functionality as demonstrated in the screenshots bellow
Screenshot 2025-10-20 at 14 38 33
Screenshot 2025-10-20 at 14 38 12

see https://docs.wagtail.org/en/stable/topics/writing_templates.html#template-fragment-caching
see #545

### DIFF
--- a/bakerydemo/templates/blog/blog_page.html
+++ b/bakerydemo/templates/blog/blog_page.html
@@ -1,38 +1,36 @@
 {% extends "base.html" %}
-{% load navigation_tags wagtailimages_tags wagtail_cache %}
+{% load navigation_tags wagtailimages_tags %}
 
 {% block content %}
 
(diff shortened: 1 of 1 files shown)


thibaudcolas (approved):
Changes for index page make sense to me! Keeping the cache tags that are most understandable by newcomers.

For the page-aware caching, will do more testing before we merge.

PR #568, opened 2025-10-23
Implement wagtailpagecache example

thibaudcolas:
Follow-up to #561 by @lupyana with the page-aware template fragment caching tag


thibaudcolas:
@lupyana relevant implementation details we were looking at:

https://github.com/wagtail/wagtail/blob/88ae925ef7bf9bab7d8f891eec09104dff064a02/wagtail/models/pages.py#L1784-L1810

Sine `last_published_at` is on there, any changes to a page that go live would naturally clear the cache. But only the changes that directly affect the page model(s). Not related models like authors or images. I’m not sure if this is a dealbreaker or no 🤔 but it might be enough of a tricky pattern that we should find a simpler example.

I’ll think a bit more about it

PR #569, opened 2025-10-25
Add screen-reader-only h2 to breads and locations index pages

### LINKED ISSUE #563: Heading hierarchy issue on breads index page
We are skipping from h1 to h3. Template:

https://github.com/wagtail/bakerydemo/blob/6a16f1b7d3c53d424361825abb9507032b5b31a4/bakerydemo/templates/breads/breads_index_page.html#L11

It seems like we could refactor this partial to use a different heading element. Or introduce a screen-reader only `h2`. Reviewing the possible solutions with @Coll76, we have decided to introduce a screen reader only `h2`. This needs to happen on two templates:

- [x] `breads_index_page`
- [x] `locations_index_page`

We already use a `u-sr-only` CSS class elsewhere for similar effect. Something like:

```html
        <h2 class="u-sr-only">{{ page.title }}</h2>
        <ul class="bread-list">
            {% for bread in breads %}
                <li>
                    {% include "includes/card/listing-card.html" with page=bread %}
                </li>
            {% endfor %}
        </ul>
```

Coll76:
Fixes #563

## Description
Fixed heading hierarchy by changing h3 to h2 in card templates to create proper h1 → h2 structure.

## Changes
* Changed `listing-card.html` heading from h3 to h2
* Changed `picture-card.html` heading from h3 to h2  
* Removed incorrect screen-reader-only h2 from `locations_index_page.html` and  `breads_index_page.html`

## Solution
Instead of adding new headings, adjusted the existing heading hierarchy to match the pattern used in `blog-listing-card.html`. Both breads and locations pages now have proper h1 → h2 hierarchy without duplicating the page title.

## Testing
* Verified heading hierarchy: h1 (page title) → h2 (individual item titles) using HeadingsMap
* No visual changes to the pages
* Follows the same pattern as the blog section
```


Scotchester (changes requested):
Thank you for your interest, however this is not a valid solution to the issue. This will cause a screen reader to repeat the page title that is already included as the `<h1>` in [header-index.html](https://github.com/wagtail/bakerydemo/blob/main/bakerydemo/templates/base/include/header-index.html#L6).

What we need to do is adjust the heading hierarchy as it currently exists, without adding new headings. This may or may not warrant modifying the styling to remain consistent with how they are currently styled.

As an aside, please add a reference to the Issue you are addressing to the PR description.

Coll76:
Thank you for the feedback! I've updated the PR to fix the heading hierarchy properly.

**Changes made:**
- Changed `listing-card.html` from h3 to h2
- Changed `picture-card.html` from h3 to h2  
- Removed the screen-reader-only h2 from `locations_index_page.html` and `breads_index_page.html`

This now creates the proper h1 → h2 hierarchy for both breads and locations pages, matching the pattern used in `blog-listing-card.html`. No new headings were added, just adjusted the existing ones as you suggested.

I've also added a reference to issue #563 in the PR description.

Issue #567, opened 2025-10-23
People section of the site


thibaudcolas:
As part of #566, we have been discussing how a new "People" section could be a really good fit for this site, for a few reasons:

- It’s a common need for Wagtail projects, so we can demonstrate something that has real-world relevance.
- It’s independent of the bread theme. This makes the demo more relevant if/when we invest into supporting other content.
- It’s relatively simple for us to produce realistic "People" content without copyright issues
- It’s a simple content type for us to have tens / possibly hundreds of items.

## Specifications

Here is what I would propose:

- A new `people` app
- With a `PeopleIndexPage` listing page
- And a `PersonPage` detail page

This would require renaming the existing `Person` model to `Author`, so it’s clear "Authors" and "People" are different content. Or alternatively, find a different name for this new section of the site.

### Landing / index page

The page could start as a copy of BreadIndexPage or LocationIndexPage. But afterwards ideally we could introduce more "landing page" content in addition to showing the `PersonPage` detail pages within the section. Perhaps the index page could use `RoutablePageMixin` to filter the people by category (say a "team" field with 5-6 different options?), but in addition opportunity to display conditional content for each of those categories?

### Detail page

Displaying a person’s profile. Ideally with content that would support showcasing the diversity of our community, so:

- Single name field, not the `first_name` and `last_name` mistake of our existing Person model
- Profile picture, with an alt text field that could say a bit more than just "Picture of |name|"
- Bio (regular StreamField?)
- Possibly a location field, either plain text or connected to location models

### Elsewhere

In addition to the above, possible related improvements:

- Feature 3-5 people on the homepage in a new section
- Possibly allow "Authors" to be connected to a "PersonPage"
- Not sure if the section would be present in the top-level menu or no (have it within "About"?)

## Working on this

Before we start PRs I think this requires more thought / feedback here in the comments, and proposed designs. Doesn’t necessarily need to be polished designs, references to existing sites / similar sections could be plenty enough. cc @lupyana who was interested to work on this

thibaudcolas:
@awais786 see the above and reflect on which aspects of this you might want to help with?

@lupyana here is what we discussed we could start with:

1. `people` app
2. `PeopleIndexPage` and `PersonDetailPage` (and templates) based on the code of the corresponding pages in the "breads" app, but with different fields for the `PersonDetailPage`
3. ??? find demo content we can reuse, in particular that we have permissions to reuse. Starting with our contributors’ profile? With public information and their permission.

[...]

PR #574, opened 2025-11-12
add people page with basic details

lupyana:
This pr adds people section with basic details to the website

used the existing breads page as base template, created people model with minimal details 

Screenshot 2025-11-12 at 17 19 59
Screenshot 2025-11-12 at 17 19 02
see #567

### DIFF
--- a/bakerydemo/people/models.py
+++ b/bakerydemo/people/models.py
@@ -2,7 +2,7 @@ from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db import models
 from wagtail.admin.panels import FieldPanel
 from wagtail.api import APIField
-from wagtail.blocks import ChoiceBlock, StructBlock, URLBlock
+from wagtail.blocks import ChoiceBlock, StructBlock, StructValue, URLBlock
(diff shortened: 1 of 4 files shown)


bakerydemo/people/models.py
```diff
@@ -0,0 +1,162 @@
...
+    image = models.ForeignKey(
+        "wagtailimages.Image",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+        help_text="Landscape mode only; horizontal width between 1000px and 3000px.",
+    )
+    body = StreamField(
+        BaseStreamBlock(), verbose_name="Page body", blank=True, use_json_field=True
```
thibaudcolas: We can go without `use_json_field`

bakerydemo/people/models.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    origin = models.ForeignKey(
```
thibaudcolas: Noting this is a confusing term in the context of people, though it definitely makes sense to reuse `Country`. Maybe change the field label here at least?

Issue #534, opened 2025-04-02
Add demo tests


thibaudcolas:
I wonder if it would be worthwhile for us to set up a few unit tests in the bakerydemo project? To demonstrate what we document is [Testing your Wagtail site](https://docs.wagtail.org/en/stable/advanced_topics/testing.html). 

For core development purposes I assume there isn’t as much of a need here, as we can just as well test those utilities in Wagtail’s test suite. But for demo purposes, writing unit tests is an important capability of Django projects, so feels worthwhile to demonstrate the application of this to Wagtail (and in particular pages and StreamField).

And should run in CI – see for example wagtail.org CI: https://github.com/wagtail/wagtail.org/blob/main/.github/workflows/ci.yml

Scotchester:
I think that'd be helpful!

Rish-it:
While we don’t need them for core Wagtail development, I think it’s still valuable to show how testing works with pages and StreamField. It should be useful for anyone using the demo as a reference.

PR #585, opened 2025-12-10
Add tests for page creation and rendering with StreamField

lupyana:
added example tests demonstrating the creation of a new child page (using a StreamField in form data) and the rendering of a page. The tests use the following assertions:

- assertCanCreate
- assertContains
- assertTemplateUsed
- assertEqual

see #534

### DIFF
--- a/bakerydemo/breads/migrations/0010_alter_breadingredient_options_alter_country_options_and_more.py
+++ b/bakerydemo/breads/migrations/0010_alter_breadingredient_options_alter_country_options_and_more.py
@@ -21,7 +21,6 @@ class Migration(migrations.Migration):
         migrations.AlterModelOptions(
             name="country",
             options={
-                "ordering": ["sort_order", "title"],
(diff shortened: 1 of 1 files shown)


bakerydemo/base/tests/test_home_page.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Verify correct template
+        self.assertTemplateUsed(response, "base/home_page.html")
+
+        # Verify content appears on the page
+        self.assertContains(response, "Welcome to our site")
+        self.assertContains(response, "Get Started")
+
+        # Alternative: Use Wagtail's built-in assertion
+        self.assertPageIsRoutable(self.home)
```
thibaudcolas: Not needed for a page type that doesn’t implement `RoutablePageMixin`

thibaudcolas (approved):
I ended up adding the CI setup to have a stronger guarantee this all works and will keep working, merging now!

PR #597, opened 2026-01-15
Rename promo_* homepage fields to lead_* for clarity

AishwaryaN483:
Previously, the homepage example fields in bakerydemo used the `promo_*` naming, which could be confused with Wagtail’s core “Promote” concepts. This change renames those fields to `lead_*` to better reflect their purpose.

Templates and fixtures were updated to match the new naming.
A new RenameField migration was added without modifying any existing migrations.
Overall behavior and UI remains unchanged.


AishwaryaN483:
Thank you!!! For merging the PR Iam going through other issues to solve and get back to the community soon..

PR #707, opened 2026-03-31
Fix hardcoded pagination condition in breads and people index templates

### LINKED ISSUE #658: Improve pagination condition in breads_index_page.html
### Issue summary

I noticed that the breads index page template had this condition:
{% if breads.paginator.count > 12 %}
this condition determines whether to show the pagination or not. But as the number is explicitly mentioned as 12 and static. It won't work correctly if the pagination size changes in the future.

Suggested improvement:
{% if breads.has_other_pages %}
this condition makes the template independent of the page size and aligns better with Django's best practices.

If this sounds good, I would be happy to submit a pull request.

### Working on this

I would like to work on this issue.

Anayo-Anyafulu:
Fixes #658.

## Changes

Replace hardcoded `{% if breads.paginator.count > 12 %}` with `{% if breads.has_other_pages %}` in `breads_index_page.html`, and the same fix in `people_index_page.html` which had the identical issue.

This makes the pagination condition independent of page size and aligns with Django best practices.

## AI usage

I used Claude (claude.ai) to help identify the fix and locate the same issue in other templates.

### DIFF
--- a/bakerydemo/templates/breads/breads_index_page.html
+++ b/bakerydemo/templates/breads/breads_index_page.html
@@ -14,7 +14,7 @@
         </ul>
     </div>
 
-    {% if breads.paginator.count > 12 %}
+    {% if breads.has_other_pages %}
         <div class="container">
             <div class="row">
                 <div class="col-sm-12">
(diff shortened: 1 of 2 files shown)


thibaudcolas (approved):
thank you!

PR #713, opened 2026-03-31
Fix missing RecipePage in search and timezone-naive is_open()

masterking008:
- Add RecipePage to native DB search fallback in search/views.py
- Replace datetime.now() with timezone.localtime() in LocationPage.is_open()

Fixes #...

### Description

- `search/views.py`: `RecipePage` was added to the project but never included in the native DB search fallback, making recipe pages invisible to search queries. Added it alongside `BlogPage`, `BreadPage`, and `LocationPage`.

- `locations/models.py`: `is_open()` was using `datetime.now()` which is timezone-naive and ignores Django's `TIME_ZONE` setting. Replaced with `timezone.localtime()` to return the correct local time.

### AI usage

Used Amazon Q Developer (AI assistant in IDE) to identify the issues and suggest fixes. Manually reviewed and verified both changes against the existing codebase before submitting.

### DIFF
--- a/bakerydemo/locations/models.py
+++ b/bakerydemo/locations/models.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from django.utils import timezone
 
 from django.conf import settings
 from django.core.validators import RegexValidator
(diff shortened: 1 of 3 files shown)


thibaudcolas:
Thank you! Note we would also need to display "Recipe" in the search results for those pages rather than "Bread". Not sure if you actually tested the search?

PR #717, opened 2026-04-05
Handle empty state in bread index page

### LINKED ISSUE #716: Handle empty state in bread index page
### Problem statement

While testing the bread index page, I noticed that when there are no breads, the page renders an empty list without any message.

This can be confusing for users, so this change adds a simple fallback message when no breads are available.

### Proposed solution

Add a conditional check before rendering the bread list.

If no breads are available, display a simple message such as “No breads available at the moment” instead of rendering an empty list.

This keeps the UI clear and avoids confusion for users when there is no content.

### Alternative solutions

_No response_

### Additional context

_No response_

### Working on this

Contribute to the discussion to help this move forward. Explain your use cases, discuss possible solutions, research how to adapt existing code, or compatibility concerns.

Once others agree with the proposed changes, anyone can contribute to this. View our [contributing guidelines](https://docs.wagtail.org/en/latest/contributing/index.html). Add a comment to the issue to explain what approach you are taking.

Sneha-Goswami123:
Fix #716 

Adds a fallback message when no breads are available on the bread index page to avoid rendering an empty list.


thibaudcolas:
thank you; I don’t see how there could ever be this situation on this site but I suppose this is a good pattern to showcase generally.

Sneha-Goswami123:
> thank you; I don’t see how there could ever be this situation on this site but I suppose this is a good pattern to showcase generally.

Thanks, that makes sense. Even if it’s not a common case here, it’s nice to have a consistent pattern.

Really appreciate the review and the improvements!

PR #724, opened 2026-04-14
Fix Person.search_fields missing SearchField for job_title

### LINKED ISSUE #723: Person.search_fields is missing SearchField for job_title — searching by job title returns no results
### Issue summary

The Person model in bakerydemo/base/models.py defines job_title only as an index.FilterField, but not as an index.SearchField. This means that when a user types a job title into the search box in the Wagtail admin People snippet, the full-text search (q=) finds no matches — even when the filter correctly narrows results.

### Steps to reproduce

Run the bakerydemo locally with load_initial_data
1) Go to Wagtail admin → Snippets → People
2) Type director into the search box and also apply the "Job title contains: director" filter
3) Observe: "No people match your query." — despite "Olivia Ava" having job_title = "Director"

### Expected Behaviour
Typing a job title like "director" in the search box should return matching people.
### Proposed Fix
Add index.SearchField("job_title") alongside the existing FilterField:
pythonsearch_fields = [
    index.SearchField("first_name"),
    index.SearchField("last_name"),
    index.SearchField("job_title"),       # ← adding this
    index.FilterField("job_title"),
    index.AutocompleteField("first_name"),
    index.AutocompleteField("last_name"),
]                                     
Both SearchField and FilterField can coexist on the same field — per [Wagtail's indexing docs](https://docs.wagtail.org/en/stable/topics/search/indexing.html): "A field can be indexed with both SearchField and FilterField at the same time."
### AI Usage
None

### Can be reproduced

Yes, on the bakerydemo

### Technical details

Python version: Python 3.12.12
Django version: 6.0.2
Wagtail version: 7.2.2
Browser version: 146

### Working on this

Yes, I would like to fix this myself. I'll add index.SearchField("job_title") alongside the existing index.FilterField("job_title") in Person.search_fields, and include a test to verify the fix.

Pra26nav:
Fixes #723 

### Description

The `Person` model in `bakerydemo/base/models.py` defined `job_title` as an `index.FilterField` but not as an `index.SearchField`. This caused full-text search (q=) in the Wagtail admin People snippet to return no results when searching by job title even though the filter worked correctly.

### AI usage

Claude (claude.ai) was used to help with git commands only. The bug was independently discovered and the fix was written manually.

### DIFF
--- a/bakerydemo/base/models.py
+++ b/bakerydemo/base/models.py
@@ -113,6 +113,7 @@ class Person(
         index.SearchField("first_name"),
         index.SearchField("last_name"),
         index.FilterField("job_title"),
+        index.SearchField("job_title"),
         index.AutocompleteField("first_name"),
         index.AutocompleteField("last_name"),
     ]


thibaudcolas:
ty!

PR #736, opened 2026-05-11
Fix TypeError in BlogIndexPage.get_child_tags sorting (#735)

SAY-5:
Closes #735.

`BlogIndexPage.get_child_tags` called `sorted(set(tags))` on `taggit.models.Tag` instances. Tag (Django model) doesn't implement `__lt__`, so any blog post with tags makes the blog index page crash with `TypeError: '<' not supported between instances of 'Tag' and 'Tag'`.

Switches to slug-based dedup (avoiding the assumption that `set(tags)` hashes by identity / slug consistently) and sorts by `name` for stable user-facing alphabetical order.

Verified the pre-fix raises `TypeError` and the post-fix returns the expected alphabetical list on a minimal Tag-shape reproducer.

### DIFF
--- a/bakerydemo/blog/models.py
+++ b/bakerydemo/blog/models.py
@@ -229,5 +229,8 @@ class BlogIndexPage(RoutablePageMixin, Page):
         for post in self.get_posts():
             # Not tags.append() because we don't want a list of lists
             tags += post.get_tags
-        tags = sorted(set(tags))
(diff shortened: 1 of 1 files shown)


laymonage (commented):
Thanks, but please add details in the issue about how you reproduced this before submitting a PR. We've been unable to reproduce the issue.

PR #737, opened 2026-05-26
Replace deprecated AccessibilityItem with ContentCheckerItem

Srishti-j18:
### Description

This PR updates `bakerydemo/base/wagtail_hooks.py` to replace the deprecated
`AccessibilityItem` userbar item with `ContentCheckerItem`.

The current implementation raises a deprecation warning on Wagtail 8.0 alpha:

> RemovedInWagtail90Warning: The userbar item 'AccessibilityItem' is deprecated. Please use 'ContentCheckerItem' instead.

This change keeps the existing behavior the same while removing that warning.

**Before:**
Screenshot 2026-05-27 at 12 32 04 AM

**After:**

Screenshot 2026-05-27 at 12 46 30 AM

### AI usage

None

### DIFF
--- a/bakerydemo/base/wagtail_hooks.py
+++ b/bakerydemo/base/wagtail_hooks.py
@@ -1,6 +1,6 @@
 from wagtail import hooks
 from wagtail.admin.filters import WagtailFilterSet
-from wagtail.admin.userbar import AccessibilityItem
+from wagtail.admin.userbar import ContentCheckerItem
 from wagtail.snippets.models import register_snippet
 from wagtail.snippets.views.snippets import SnippetViewSet, SnippetViewSetGroup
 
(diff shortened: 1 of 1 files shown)


bakerydemo/base/wagtail_hooks.py
```diff
@@ -1,6 +1,6 @@
 from wagtail import hooks
 from wagtail.admin.filters import WagtailFilterSet
-from wagtail.admin.userbar import AccessibilityItem
+from wagtail.admin.userbar import ContentCheckerItem
```
Srishti-j18, line 3: bakerydemo CI is still using `wagtail>=7.2,<7.3`, which is why CI is raising an ImportError:
https://github.com/wagtail/bakerydemo/blob/2a2f0f8c3525797f8066645659f02f28ab12b2a4/requirements/base.txt#L3

We can either update `requirements/base.txt` to a newer Wagtail version, or use a try/except import so we avoid the deprecation warning while remaining compatible with the current supported version.
```suggestion
try:
    from wagtail.admin.userbar import ContentCheckerItem as  ContentCheckerItem
except  #ImportError:
    from wagtail.admin.userbar import AccessibilityItem as ContentCheckerItem

```

Srishti-j18: but also it seems like a workaround rather than the right fix. It would be better either to keep the current implementation or to bump the supported Wagtail version before switching to `ContentCheckerItem`.

PR #748, opened 2026-06-04
Add uv support for local development, CI, and Docker

Srishti-j18:
Task of #739 

### Description

This adds uv support to bakerydemo while keeping the existing requirements-based workflow and manual venv setup supported.

**Tested locally**
I tested both the direct local uv setup and the local Docker setup.

### AI usage

I used GPT to help with the changes, and I manually reviewed the changes and verified them locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,10 @@
+ARG UV_VERSION=0.11.16
+FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv
+
 FROM python:3.12-slim
 ARG NIGHTLY=0
+
+COPY --from=uv /uv /uvx /bin/
 
 # Install packages needed to run your application (not build deps):
 # We need to recreate the /usr/share/man/man{1..8} directories first because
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,11 @@
+ARG UV_VERSION=0.11.16
+FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv
+
 FROM python:3.12-slim
```
thibaudcolas, line 4: Need to double check if we should be doing this like that with a multi-stage build, or try to use one slim image with uv preinstalled. Or install uv another way

readme.md
```diff
@@ -165,6 +166,64 @@ To set up your database and load initial data, run the following commands:
 
 Log into the admin with the credentials `admin / changeme`.
 
+## Setup with uv
```
thibaudcolas, line 169: Need to double check if we could make this shorter?

thibaudcolas (commented):
Looking good! Needs more review

